PR #24, opened 2025-10-12
Peng depositing funds

pengying:


mintlify/snippets/depositing-funds.mdx
```diff
@@ -0,0 +1,86 @@
+Grid provides two options to fund an account:
+
+- Prefund
+- Just-in-time funding
+
+With prefunding, you'll deposit funds into internal accounts via Wire, PIX, or crypto transfers. You can then use the balances as the source of funds for quotes and transfers.
```
jklein24, line 6: Why are we saying "wire, pix, or crypto" as the funding types? You can't actually fund an internal account over PIX at the moment and we should definitely mention ACH, etc. I'm not sure we should be specific here about bank rails at all. Maybe we can just say "via bank or crypto transfers"

mintlify/docs.json
```diff
@@ -25,7 +25,7 @@
         "tab": "Payouts",
         "pages": [
           "payouts/index",
-          "payouts/terminology",
+          "developer-resources/terminology",
```
jklein24, line 28: @pengying it seems like you may have broken all of the links to these. I'll fix them, but just a heads up...

mintlify/snippets/depositing-funds.mdx
```diff
@@ -0,0 +1,86 @@
...
+
+</CodeGroup>
+</Step>
+
+<Step title="Extract payment instructions">
+Parse the `fundingPaymentInstructions` array and select the appropriate instructions based on your customer's preferred payment method.
+
+```javascript
+const instructions = account.fundingPaymentInstructions[0];
+const bankInfo = instructions.bankAccountInfo;
```
jklein24, line 57: Random note I just noticed... If we're going to allow crypto funding, we should probably rename this to accountInfo to remove the "bank", right?

jklein24:
Is this not going to have a bunch of duplicated info with the existing internal accounts guides? Might be good to dedupe where it makes sense.

PR #32, opened 2025-10-12
Peng country support holder

pengying:


mintlify/global-p2p/country-support.mdx
```diff
@@ -0,0 +1,17 @@
+---
+title: "Supported Countries"
+---
+
+## Customer onboarding (non regulated)
+- US
+- EU
+- MX
```
jklein24, line 8: I don't think we can actually do this, can we? Doesn't matter a ton, but I think US and EU seems like all we can do ourselves for now.

mintlify/global-p2p/country-support.mdx
```diff
@@ -0,0 +1,17 @@
...
+
+## Sending and Receiving Payments
+### Country & Banking Rail
```
jklein24, line 11: We should probably include the projected 64 countries by EOY like David mentioned. This would include all of the taza countries, etc.

We also might want to note our crypto support.

mintlify/global-p2p/country-support.mdx
```diff
@@ -0,0 +1,17 @@
+---
+title: "Supported Countries"
```
jklein24, line 2: We should probably reconcile this with the existing rails-by-country.mdx file. I'm not sure that one is correct or complete, but we should align to whatever is correct and only write it out once.

PR #78, opened 2025-10-15
Update Plaid external account linking to return external account object

pengying:


mintlify/snippets/plaid-integration.mdx
```diff
@@ -245,8 +245,31 @@ app.post('/api/plaid/exchange-token', async (req, res) => {
 
 ```json
 {
-  "message": "External account creation initiated. You will receive a webhook notification when complete.",
-  "requestId": "req_def456ghi789"
+  "id": "ExternalAccount:a12dcbd6-dced-4ec4-b756-3c3a9ea3d123",
```
whoisglover, line 248: This wont actually return all this data synchronously. Its an async job that will send a webhook once its done

pengying: Yeah I know. I can update the example to exclude account info.

pengying: hmm I made accountInfo option for this, but we may just want to split it out into 2 objects

PR #88, opened 2025-11-02
Add Lightning as a funding and external account option

jklein24:
1. Adding a lightning invoice as an option for real-time funding of a quote, which is useful for offramps.
2. Add an external account type for lightning which can use an invoice, bolt12 offer, or lightning address for offramps.


jklein24:
* **#88** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/88?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/88?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jklein24:
The one thing I want to think through a bit more on this PR is internal account funding. For JIT quote funding, it makes sense to only use a lightning invoice, but for funding internal accounts, it's much messier to have a static funding address. However, I think for now, because we don't have BTC internal accounts, we would just never have lightning `fundingPaymentInstructions` for an internal account. This makes me comfortable kicking that can down the road. Maybe bolt12 will catch on more before we need to support that use case.

exs6350:
Have to re-run the generator. Is this lightning something that we want to also support by end of year or something we can take on next quarter?

jklein24:
> Have to re-run the generator. Is this lightning something that we want to also support by end of year or something we can take on next quarter?

Good question on timing. I'm not totally sure tbh... I do think that lightning is probably critical for btc on/off ramps today. @whoisglover I'm curious whether you see this as adding significant scope on the LSP side.

openapi/components/schemas/external_accounts/LightningExternalAccountInfo.yaml
```diff
@@ -0,0 +1,37 @@
+allOf:
+  - type: object
+    required:
+      - accountType
+    properties:
+      accountType:
+        type: string
+        enum: [LIGHTNING]
+        example: LIGHTNING
+  - oneOf:
```
pengying, line 10: Might need a discriminator to deserialize in the SDK.

pengying: Could probably yolo it and I can clean it up for SDK gen later

openapi/components/schemas/external_accounts/LightningExternalAccountInfo.yaml
```diff
@@ -0,0 +1,37 @@
...
+      - title: Lightning Invoice
+        type: object
+        required:
+          - invoice
+        properties:
+          invoice:
+            type: string
+            description: 1-time use lightning bolt11 invoice payout destination
+            example: lnbc15u1p3xnhl2pp5jptserfk3zk4qy42tlucycrfwxhydvlemu9pqr93tuzlv9cc7g3sdqsvfhkcap3xyhx7un8cqzpgxqzjcsp5f8c52y2stc300gl6s4xswtjpc37hrnnr3c9wvtgjfuvqmpm35evq9qyyssqy4lgd8tj637qcjp05rdpxxykjenthxftej7a2zzmwrmrl70fyj9hvj0rewhzj7jfyuwkwcg9g2jpwtk3wkjtwnkdks84hsnu8xps5vsq4gj5hs
```
pengying, line 19: Know that's it's also referenced in quote.destination external account but feels kind of weird to have a 1 time use invoice persisted as an external account.  Maybe makes sense to separate them at some point down the road.

jklein24: Yeah, agreed :-/... Wasn't totally sure the best way to represent it otherwise because we'd need to move the invoice up a level to payment instructions only or something... Need to think that through a bit.

whoisglover (approved):

whoisglover (approved):

whoisglover (approved):

PR #95, opened 2025-11-21
[Grid] Remove add platform external account endpoint

exs6350:
Given that a platform will be its own first customer, this endpoint is redundant because a platform can use the /customers/external-accounts endpoint and not pass a customer id (or they can) to onboard an external account for themselves.


exs6350:
* **#95** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/95?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/95?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jklein24 (commented):
Sorry, I know we went back and forth on this quite a bit, but I'd thought we'd landed on the decision that including sends from the platform was the better devX, cleaner in our docs, etc. Can you remind me of the full result of those conversations?

exs6350:
https://docs.google.com/document/d/15rZ7dPErwLC1FLxgxfmAB_2TSidReBoG1Jo7Nerx2vw/edit?tab=t.baeiuqv2ttv9#bookmark=id.xmus9do1dr0u We decided that the platform themselves would be their own first customer which would make it easy to do platform level payouts. In terms of devx we still want platform level apis but this specific endpoint is just a 1:1 duplicate of /customers/external-account. I think maybe the better approach here is to make a new input object for the platform and keep this endpoint

jklein24:
> https://docs.google.com/document/d/15rZ7dPErwLC1FLxgxfmAB_2TSidReBoG1Jo7Nerx2vw/edit?tab=t.baeiuqv2ttv9#bookmark=id.xmus9do1dr0u We decided that the platform themselves would be their own first customer which would make it easy to do platform level payouts. In terms of devx we still want platform level apis but this specific endpoint is just a 1:1 duplicate of /customers/external-account. I think maybe the better approach here is to make a new input object for the platform and keep this endpoint

Got it, hooray meeting notes! So the plan is to start by automatically creating a customer when the platform is created, then we'll migrate to doing this correct platform handling later? FWIW I'm slightly worried that this migration will be painful (move all transactions, balances, etc), but if it's lots of extra work to do the right thing now, I agree with the gradual approach. What should we do with the docs, though?

exs6350:
I think we just need to add a section to the docs to clarify that the platform themselves will be onboarded as a customer. So any payouts, btc rewards, etc can be done with any of the existing customer apis.

I don't think there's any extra work for us to do in handling a platform. We added a `EntGridAccountOwnerPattern` that could be either a platform or customer so this is just added whenever an object can be owned by either one

PR #97, opened 2025-11-26
[Grid] Remove required currency from uma source

exs6350:
When we onboard accounts we have a good idea what currency they are backed in however thats not always the case with a uma transaction.

For example, in a receive uma to uma transaction we would not know what the originating currency is so we shouldn't make currency required for a uma source but we can make it required for an account source

### DIFF
--- a/openapi/components/schemas/transactions/Transaction.yaml
+++ b/openapi/components/schemas/transactions/Transaction.yaml
@@ -22,6 +22,7 @@ properties:
         type: object
         required:
           - accountId
+          - currency
         properties:
           accountId:
             type: string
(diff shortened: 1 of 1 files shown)


exs6350:
* **#97** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/97?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/97?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jklein24 (approved):

PR #115, opened 2025-12-17
[Grid] Move source into outgoing/incoming transaction

exs6350:
This moves source out of the base transaction object and into the incoming/outgoing transaction. We still enforce it for outgoing transaction but not for a incoming transaction. The reason why we need to do this is because an incoming transaction like a funding event will not have a source

### DIFF
--- a/mintlify/openapi.yaml
+++ b/mintlify/openapi.yaml
@@ -5627,7 +5627,6 @@ components:
         - id
         - status
         - type
-        - source
         - destination
         - customerId
         - platformCustomerId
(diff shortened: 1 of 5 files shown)


exs6350:
* **#115** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/115?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/115?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

pengying (approved):
oneofs usually need a discriminator so long term this will likely need to change but for right now should be ok.

jklein24 (approved):

PR #140, opened 2026-01-28
fix: reverting back to using enums instead of const for openapi generator

pengying:
### TL;DR

Replace `const` with `enum` arrays in OpenAPI schema definitions for better compatibility with OpenAPI tooling.

### What changed?

This PR replaces all instances of the `const` keyword with `enum` arrays containing a single value throughout the OpenAPI schema definitions. For example:

```diff
- const: 401
+ type: integer
+ enum:
+   - 401
```

The change affects various schema components including:
- HTTP status codes in error responses
- Account type identifiers
- Customer types
- Asset types
- Webhook event types


pengying:
* **#140** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/140?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/140?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

openapi/components/schemas/common/PaymentSolanaWalletInfo.yaml
```diff
@@ -1,9 +1,11 @@
...
-      - USDT
\ No newline at end of file
+  - type: object
+    properties:
+      assetType:
+        type: string
+        description: Type of asset
+        enum:
+          - USDC
+          - USDT
```
gcamata, line 11: shouldn't this be a ref?

pengying: We could make it a ref if you want to create an enum for openapi-generator for webdev.

```python
    @field_validator('asset_type')

    def asset_type_validate_enum(cls, value):
        """Validates the enum"""
        if value is None:
            return value
        if value not in set(['USDC', 'USDT']):
            raise ValueError("must be one of enum values ('USDC', 'USDT')")
        return value
    model_config = ConfigDict(
        populate_by_name=True,
        validate_assignment=True,
        protected_namespaces=(),
    )
```

For external generation via stainless, this results in a distinct type:  
https://github.com/stainless-sdks/grid-kotlin/blob/main/grid-kotlin-core/src/main/kotlin/com/grid/api/models/quotes/PaymentInstructions.kt#L4432

```kotlin
class AssetType @JsonCreator private constructor(private val value: JsonField<String>) :
                Enum {

                /**
                 * Returns this class instance's raw value.
                 *
                 * This is usually only useful if this instance was deserialized from data that
                 * doesn't match any known member, and you want to know that value. For example, if
                 * the SDK is on an older version than the API, then the API may respond with new
                 * members that the SDK is unaware of.
                 */
                @com.fasterxml.jackson.annotation.JsonValue fun _value(): JsonField<String> = value

                companion object {

                    val USDC = of("USDC")

                    val USDT = of("USDT")

                    fun of(value: String) = AssetType(JsonField.of(value))
                }

                /** An enum containing [AssetType]'s known values. */
                enum class Known {
                    USDC,
                    USDT,
                }

```

gcamata (approved):

PR #222, opened 2026-02-23
FX Rate Explorer: interactive rate comparison table

patcapulong:
## Summary

Interactive FX rate comparison page at `/fx-rates` that lets visitors compare Grid's pricing against major competitors (Stripe, Airwallex, Wise, PayPal, Remitly, Western Union, OFX, and others) across 40+ currency corridors.

### Data sources
- **Grid column**: Live mid-market rates from Grid's public exchange-rates API, with Grid's published spread applied per corridor
- **Wise + dynamic providers**: Live quotes from Wise's public comparison API (`/v4/comparisons/`), which returns data for Wise, PayPal, Remitly, Western Union, banks, and others
- **Stripe & Airwallex**: Computed from publicly documented pricing schedules (FX markup + cross-border fees + fixed fees)
- **Bank avg.**: Average of bank-type providers returned by the Wise comparison API

### Key features
- **40+ currencies** with flag icons (circle-flags CDN), organized by region
- **Custom searchable dropdowns** for source and destination currency selection, with keyboard navigation and scroll-aware gradient masks
- **Currency chip filters** — select specific destinations or view all corridors
- **Two view modes**: Cost in basis points (bps) vs. Amount received
- **Best-in-row highlighting** — cheapest provider per corridor gets green text (including competitors, for credibility)
- **Delta column** — shows how much better/worse Grid is vs. the best competitor
- **Sticky header** with scroll-direction-aware hide/show (hides on scroll down, reappears on scroll up)
- **Column hover** — hovering a cell highlights the column header
- **Skeleton loading** with static placeholder for zero layout shift
- **"Get custom pricing" banner** linking to contact sales

### Performance
- **Batched API requests**: 10 concurrent, sequential batches to avoid hammering Wise API
- **sessionStorage caching**: 30-min fresh TTL, 2-hour stale-while-revalidate
- **Amount bucketing**: rounds send amounts to reduce cache key cardinality
- **Viewport-gated loading**: IntersectionObserver defers all fetches until component is near viewport
- **1200ms debounce** on amount input changes

### Files changed
- `mintlify/fx-rates.mdx` — entire component (data, logic, JSX) in a single MDX file
- `mintlify/style.css` — all `rate-explorer-*` and `fx-rates-*` CSS classes
- `mintlify/docs.json` — nav entry updated to `fx-rates`
- `mintlify/images/icons/` — competitor logos (Stripe, Wise, PayPal, Remitly, WU, OFX, Airwallex, etc.)
- `mintlify/images/og/og-fx-rates.png` — dedicated Open Graph image

## Test plan
- [x] Page loads at `/fx-rates` with wide layout
- [x] Skeleton renders immediately, then live data populates after ~1-2s
- [x] Source currency dropdown: search, keyboard nav, "Coming soon" section for unsupported currencies
- [x] Destination currency: select → chips appear, table filters; "+ Add" for more; "Clear" resets to all
- [x] Amount input: comma formatting, debounced re-fetch on change
- [x] Toggle between Cost (bps) and Amount received — values update correctly
- [x] Best provider per row highlighted in green
- [x] Delta column shows +/- vs best competitor
- [ ] Sticky header hides on scroll down, reappears on scroll up
- [ ] Column hover highlights the header
- [ ] Dark mode: all colors flip correctly
- [ ] Mobile: header stacks, table scrolls horizontally
- [ ] "Get custom pricing" banner renders with background image and hover effect
- [ ] OG image renders when sharing the URL


mintlify/fx-rates.mdx
```diff
@@ -0,0 +1,827 @@
...
+  { code: 'MXN', name: 'Mexican Peso', region: 'Americas', usdMid: 17.45, gridSpread: 0.5, wiseSpread: 0.95 },
+  { code: 'BRL', name: 'Brazilian Real', region: 'Americas', usdMid: 5.05, gridSpread: 0.55, wiseSpread: 1.00 },
+  { code: 'CRC', name: 'Costa Rican Colón', region: 'Americas', usdMid: 512, gridSpread: 0.8, wiseSpread: 1.60 },
+  { code: 'EUR', name: 'Euro', region: 'Europe', usdMid: 0.921, gridSpread: 0.25, wiseSpread: 0.55 },
+  { code: 'GBP', name: 'British Pound', region: 'Europe', usdMid: 0.791, gridSpread: 0.25, wiseSpread: 0.55 },
+  { code: 'CHF', name: 'Swiss Franc', region: 'Europe', usdMid: 0.881, gridSpread: 0.3, wiseSpread: 0.55 },
+  { code: 'DKK', name: 'Danish Krone', region: 'Europe', usdMid: 6.87, gridSpread: 0.3, wiseSpread: 0.55 },
+  { code: 'SEK', name: 'Swedish Krona', region: 'Europe', usdMid: 10.45, gridSpread: 0.4, wiseSpread: 0.90 },
+  { code: 'NOK', name: 'Norwegian Krone', region: 'Europe', usdMid: 10.55, gridSpread: 0.4, wiseSpread: 0.90 },
+  { code: 'CZK', name: 'Czech Koruna', region: 'Europe', usdMid: 23.15, gridSpread: 0.45, wiseSpread: 0.95 },
```
pengying: do we need to hardcode these mid and spread values?

mintlify/fx-rates.mdx
```diff
@@ -0,0 +1,827 @@
...
+  GHS: 'gh', UGX: 'ug', TZS: 'tz', ZMW: 'zm', MWK: 'mw',
+  XOF: 'sn', XAF: 'cm', CDF: 'cd', BWP: 'bw',
+  INR: 'in', PHP: 'ph', IDR: 'id', SGD: 'sg', HKD: 'hk',
+  CNY: 'cn', KRW: 'kr', THB: 'th', VND: 'vn', MYR: 'my', LKR: 'lk',
+};
+
+export const flagUrl = (code) =>
+  'https://hatscripts.github.io/circle-flags/flags/' + (CURRENCY_TO_COUNTRY[code] || code.slice(0, 2).toLowerCase()) + '.svg';
+
+export const SOURCE_CODES = ['USD', 'EUR', 'GBP', 'CAD', 'SGD', 'HKD'];
```
pengying, line 66: Should we start with USD, USDC?  We don't support the other source currencies yet.

mintlify/fx-rates.mdx
```diff
@@ -0,0 +1,827 @@
...
+};
+
+export const WISE_API_BASE = 'https://api.transferwise.com/v4/comparisons/';
+
+export const CACHE_TTL = 30 * 60 * 1000;
+export const STALE_TTL = 2 * 60 * 60 * 1000;
+export const BATCH_SIZE = 10;
+
+export const getStripeFxBps = (source, dest) => {
+  const majorPairs = ['USD-EUR','USD-GBP','EUR-USD','GBP-USD','EUR-GBP','GBP-EUR'];
```
pengying, line 100: where are you getting the stripe exchange rates?

mintlify/fx-rates.mdx
```diff
@@ -0,0 +1,827 @@
...
+                              <span className="rate-explorer-provider-amount">{formatNumber(provider.receivedAmount, recDec)} {row.code}</span>
+                            )
+                          ) : (
+                            <span className="rate-explorer-na">{'\u2014'}</span>
+                          )}
+                        </td>
+                      );
+                    })}
+                    <td className={'rate-explorer-provider-cell' + (bestKey === 'BankAvg' ? ' rate-explorer-provider-best' : '')}>
+                      {row.bankAvgBps !== null ? (
```
pengying, line 1068: instead of bank average, can we pick a bank either Chase or BoA?

mintlify/fx-rates.mdx
```diff
@@ -0,0 +1,827 @@
...
+
+export const getStripeTotalBps = (source, dest, sendAmount) => {
+  const fxBps = getStripeFxBps(source, dest);
+  const cbBps = STRIPE_CB_BPS[dest] || 100;
+  const fixedBps = sendAmount > 0 ? Math.round(STRIPE_FIXED_FEE / sendAmount * 10000) : 0;
+  return fxBps + cbBps + fixedBps;
+};
+
+export const AIRWALLEX_MAJORS = ['USD','EUR','GBP','JPY','CHF','AUD','NZD','CAD','HKD','SGD','CNY'];
+export const getAirwallexBps = (dest) => AIRWALLEX_MAJORS.includes(dest) ? 50 : 100;
```
pengying, line 123: not sure if you want to get claude to make all of the requests but airwallex does provide a fx api https://www.airwallex.com/api/fx/fxRate/liveRate?sellCcy=USD&buyCcy=EUR

mintlify/fx-rates.mdx
```diff
@@ -0,0 +1,827 @@
...
+export const BATCH_SIZE = 10;
+
+export const getStripeFxBps = (source, dest) => {
+  const majorPairs = ['USD-EUR','USD-GBP','EUR-USD','GBP-USD','EUR-GBP','GBP-EUR'];
+  if (majorPairs.includes(source + '-' + dest)) return 50;
+  if (source === 'USD') return 100;
+  return 200;
+};
+
+export const STRIPE_CB_BPS = {
```
pengying, line 106: maybe we can hide bridge from the amount received table since we can't get an accurate fx rate

mintlify/fx-rates.mdx
patcapulong: claude told me this 🫤 

> The Airwallex API is just a mid-market rate feed that powers their website converter. It's the "before fees" number they show to attract customers. The real 50 bps markup gets added when you actually send money — it's not in the API response at all.
> 
> We proved this by hitting the API in both directions (sell USD→buy EUR, then sell EUR→buy USD) at the same instant. If 50 bps were baked in, you'd see a ~100 bps gap between the two. Instead, the gap is only ~3 bps — interbank-level pricing. The 50 bps lives in their transaction layer, not their display rate API.
> 
> So using it would give us: near_interbank_rate + 50bps — which is the same thing we already compute as grid_mid_rate - 50bps. Same answer, more complexity, and only 10 currencies instead of our 40+.

mintlify/fx-rates.mdx
patcapulong: i'm sorry, but i have another response i had claude write up for me. idk how to say it better.

the question i asked is like, "should we hide the stripe amount received values because we don't know their FX rate, only their fees?"

> Stripe's Global Payouts pricing page lists the FX cost as "0.50% of the amount" for USD/EUR/GBP conversions, plus a cross-border fee per corridor (e.g., 25 bps to UK/EU). We apply these published fees to the mid-market rate, same as the mid-market baseline the table already uses.
>
>Could Stripe have an additional hidden spread in their base rate beyond the published fee? It's possible but unlikely — the fee is presented as the total FX cost, and undisclosed FX markups would be a regulatory issue (PSD2 requires full cost disclosure in the EU, and Stripe is publicly traded). We can't independently verify without a test account, but the same is true for any provider in the table — the bank data from the Wise Comparison API is also third-party estimates that can be hours stale.
>
>The approach is the same one Wise themselves use on their comparison page for providers they don't have direct data from. We already updated the footer to explain the methodology: "Stripe and Airwallex amounts estimated from published fee schedules applied to the mid-market rate" with links to their pricing docs.
>
>(Source: https://docs.stripe.com/global-payouts/pricing)

i know we want rigor, but i think this is fine for this use case. we are just listing amounts for user convenience and explain it in the footer. happy to chat about this

mintlify/fx-rates.mdx
patcapulong: > We don't have Stripe's actual exchange rate — they don't publish it or expose an API for it. What we have is their published fee structure (FX fee + cross-border fee + fixed fee per corridor) from https://docs.stripe.com/global-payouts/pricing. We apply those fees to the mid-market rate to estimate what a customer would receive.
> 
> Addressed this in more detail on the other thread — short version is this is the same approach Wise uses on their comparison page for providers without live rate data, and we explain the methodology in the footer.

LMK what you think. i'm happy to implement the thing where we hide stripe data on the "amount received" view, i think it's just better to have the table filled ya kno

mintlify/fx-rates.mdx
pengying: can you ask claude in that context - if Stripe / Bridge is converting from a stablecoin sandwich converting from fiat to USDC and USDC to local fiat instead of banking partners, would it still follow the mid market rate?

mintlify/fx-rates.mdx
pengying: where is this getting the mid market rate btw?

pengying: Or is it assuming that our rates are mid market rates?

pengying: airwallex docs also say that they compare multiple FX liquidity providers https://www.airwallex.com/docs/transactional-fx/get-started/choose-your-fx-solution/rates

mintlify/fx-rates.mdx
patcapulong: ya 😆 📞 

> Good question. The stablecoin sandwich is USD → USDC (on-ramp at ~1:1, Circle redeems at par) → USDC → local fiat (off-ramp). Since USDC is a USD-pegged asset, the USDC/EUR rate is economically equivalent to USD/EUR — arbitrageurs keep these aligned within a few bps for major pairs.
> 
> But more importantly: even if the stablecoin route introduces a few bps of friction vs traditional FX, that's Stripe's internal cost, not the customer's. Stripe publishes a fixed fee schedule ("0.50% of the amount") — they absorb any routing inefficiency from their own margin. The stablecoin sandwich is why they acquired Bridge: it's cheaper than correspondent banking, which means better margins for Stripe at the same published price to the customer.
> 
> So yes, it should track mid-market very closely, and any small deviation would be on Stripe's dime, not the customer's.

mintlify/fx-rates.mdx
patcapulong: > We dug into this pretty thoroughly. Here's what we found:
> 
> Airwallex's public API (/liveRate) is a display rate feed — it powers their website currency converter. The rates it returns are near-interbank (~3 bps spread total when you compare forward vs reverse directions). The 50 bps markup from their pricing page gets applied at transaction time, not in the API response.
> 
> Their docs confirm this two-layer model: a "proprietary Rates pricing engine" that blends quotes from multiple bank liquidity providers (dynamic, based on volatility/size/liquidity), and then "any additional agreed margin applied on top" — which is the 50 bps for major currencies.
> 
> So even if we called the API, we'd get the pre-margin rate and still need to add 50 bps on top. Net result is basically the same as our current approach (mid-market - 50 bps).
> 
> We looked into using it anyway for the Airwallex column specifically, but there are two hard blockers:
> 1. No CORS headers — browser blocks cross-origin requests to their API, and we're running client-side in Mintlify (no server component to proxy through)
> 2. Only 11 currencies supported (USD, EUR, GBP, JPY, CHF, AUD, NZD, CAD, HKD, SGD, CNY) — everything else returns 400. We'd still need the hardcoded fallback for the other 30+ currencies in our table.
> 
> For ~2-5 bps of additional accuracy on 11 out of 40+ rows, standing up a proxy doesn't seem worth it for a docs page. The hardcoded 50/100 bps from their published pricing is actually slightly generous to Airwallex (real total cost is probably 52-55 bps once you include the engine-level spread).
> 
> Source: https://www.airwallex.com/docs/transactional-fx/get-started/choose-your-fx-solution/rates

mintlify/fx-rates.mdx
patcapulong: yeah i put BoA because it has 10 corridors. chase only has 5. we can do either, i already have the assets and stuff. LMK what you think.

mintlify/fx-rates.mdx
patcapulong: is it USD/USDC? or you can pick USD or USDC. we can't send EUR? or GBP? idk. it feels sad to only have  USD lol. i thought audience for this is also prospective customers — wouldn't we be limiting ourselves a lot?

patcapulong: for now i just have it use the same dropdown as the destination currency, and the unavailable currencies i say "coming soon"

mintlify/fx-rates.mdx
patcapulong: > Yes — these are last-resort fallback values, only used when all live data sources fail. The table has a 4-tier fallback cascade:
> 
> 1. **Fresh cache** (sessionStorage, < 30 min) — immediate, no fetch
> 2. **Stale cache** (sessionStorage, 30 min–2 hr) — shown immediately while background refetch happens
> 3. **Last-known-good** (localStorage, any age) — persists across sessions from a previous successful fetch
> 4. **Hardcoded array** — static snapshot, only hit if the user has never had a successful fetch
> 
> In practice, tiers 1–3 cover almost all cases. A user would need to visit for the first time ever *and* have both the Grid and Wise APIs down simultaneously to see hardcoded values. They're dated (2026-02-24) so we know when to refresh them.

mintlify/fx-rates.mdx
patcapulong: i reverted this because i accidentally reverted to last commit and lost all the work for this like a dummy. it was a lot to specify which bank to put for each sending currency and whatever. i kinda think bank average is fine. people looking at this should know that banks are not competitive at all

patcapulong:
@greptile can you review this

PR #236, opened 2026-02-27
feat: add cryptoNetwork field to RealtimeFundingQuoteSource

kphurley7:
## Summary

Adds an optional `cryptoNetwork` field to `RealtimeFundingQuoteSource`.

Callers depositing stablecoins (USDC, USDT) via realtime funding need to specify which network they are depositing on so the server can generate a deposit address on the correct network. Without this field, the server has to hardcode the network (e.g. Solana devnet for sandbox), which won't work when production supports multiple networks.

The field is **optional** for backward compatibility — existing fiat realtime funding sources (USD, MXN, etc.) are unaffected.

## Usage

```json
{
  "source": {
    "sourceType": "REALTIME_FUNDING",
    "currency": "USDC",
    "cryptoNetwork": "SOLANA_MAINNET",
    "customerId": "Customer:..."
  }
}
```

## Test plan

- [ ] Confirm existing fiat realtime funding quote tests still pass (no `cryptoNetwork` required)
- [ ] Confirm USDC realtime funding quote with `cryptoNetwork: "SOLANA_MAINNET"` returns a Solana deposit address
- [ ] OpenAPI lint passes (`make lint`)

### DIFF
--- a/openapi/components/schemas/quotes/RealtimeFundingQuoteSource.yaml
+++ b/openapi/components/schemas/quotes/RealtimeFundingQuoteSource.yaml
@@ -23,6 +23,14 @@ allOf:
           [Supported Currencies](https://grid.lightspark.com/platform-overview/core-concepts/currencies-and-rails)
           for the full list of supported fiat and crypto currencies.
         example: USD
+      cryptoNetwork:
+        type: string
+        description: >-
+          The crypto network to use for the funding source. Required when `currency` is a
+          stablecoin (e.g. USDC, USDT). Specifies which network the customer will deposit
+          on, so the correct deposit address can be generated. Example values:
+          `SOLANA_MAINNET`, `SOLANA_DEVNET`, `ETHEREUM_MAINNET`.
+        example: SOLANA_MAINNET
     description: >-
       Fund the quote using a real-time funding source (RTP, SEPA Instant, Spark, Stables, etc.).
       This will require manual just-in-time funding using `paymentInstructions` in the response.
--- a/mintlify/openapi.yaml
+++ b/mintlify/openapi.yaml
@@ -7751,6 +7751,10 @@ components:
               type: string
               description: Currency code for the funding source. See [Supported Currencies](https://grid.lightspark.com/platform-overview/core-concepts/currencies-and-rails) for the full list of supported fiat and crypto currencies.
               example: USD
+            cryptoNetwork:
+              type: string
+              description: 'The crypto network to use for the funding source. Required when `currency` is a stablecoin (e.g. USDC, USDT). Specifies which network the customer will deposit on, so the correct deposit address can be generated. Example values: `SOLANA_MAINNET`, `SOLANA_DEVNET`, `ETHEREUM_MAINNET`.'
+              example: SOLANA_MAINNET
           description: Fund the quote using a real-time funding source (RTP, SEPA Instant, Spark, Stables, etc.). This will require manual just-in-time funding using `paymentInstructions` in the response. Because quotes expire quickly, this option is only valid for instant payment methods. Do not try to fund a quote with a non-instant payment method (ACH, etc.).
     QuoteSourceOneOf:
       oneOf:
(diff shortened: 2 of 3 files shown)


openapi/components/schemas/quotes/RealtimeFundingQuoteSource.yaml
```diff
@@ -23,6 +23,14 @@ allOf:
           [Supported Currencies](https://grid.lightspark.com/platform-overview/core-concepts/currencies-and-rails)
           for the full list of supported fiat and crypto currencies.
         example: USD
+      cryptoNetwork:
+        type: string
```
pengying, line 27: @wuvictor-95 I thought we were going to return an array of all supported networks in the quote response payment instructions?

wuvictor-95: It's what we are doing currently since we only have one network. In the future though, as we expand our capabilities to many networks, it makes less sense to do. We wouldn't want to create 3+ garbage deposit addresses for different networks everytime we create a quote

pengying: Oh man.  This is another breaking change.  And I thought address generation was cheap and we were going to use that pattern for other payment instructions.

wuvictor-95: This could be handled via config e.g. platform specifies which networks they care about when funding payments. Default is solana. 

Yea, it's cheap but not 0. We have to check in each address to be watched on Alchemy, and alchemy has limits. It's definitely not scalable to, say, spin up 5 addresses for every payment. But perhaps we can manage that complexity and clean up all the addresses after 24 hours.

pengying: It's better to get breaking changes in now.  I do think it's a better experience to be able to accept across multiple chains.  As an example if I'm depositing to Uniswap, I generate the address once and can send to it from any network.  Vs needing to specify the network prior to address generation.

wuvictor-95: Yea, I think it really depends on the use case. I can totally see a platform who can fulfill payments on behalf of users, and gives the user multiple options on how to fund the payment. 

It's possible to manage the complexity on our side and always present multiple deposit addresses. I'm ok with leaving as is if you see viable use cases.

jklein24: This isn't actually technically a breaking change though from an API standpoint since we're adding a new optional field. We could just return all available networks if no network is specified, right? Or we could just error out and then the like 3 people testing sandbox will see it and we can flag that we're adding this in sandbox. Not a huge deal imo.

pengying: It's not a breaking change if it's totally optional and we return all configured networks even when the currency is a stablecoin.    
  
The description below makes it seem required when currency is any stablecoin

jklein24: We should make a call here. IMO, we should just merge this and in the short term if it's empty we can return all available (we only support solana right now anyway). I want to unblock @kphurley7 to merge this.

jklein24 (approved):

jklein24 (approved):

openapi/components/schemas/quotes/RealtimeFundingQuoteSource.yaml
jklein24: We should make a call here. IMO, we should just merge this and in the short term if it's empty we can return all available (we only support solana right now anyway). I want to unblock @kphurley7 to merge this.

pengying: yeah, based on victor's response I thought we were going to make this optional even for crypto currencies and return all available?  That decision makes sense to me

wuvictor-95: Right now we are returning a list of payment instructions.

The PR description seems like this was auto-created since we're hardcoding to create a solana address in paycore. This is intended. We will hardcode networks in the backend since the original strategy was to just return a list of networks anyways.

I think we can close this PR. If we want to add CryptoNetwork as a optional field later on, we can. Not sure what work this is blocking though.

jklein24 (approved):

PR #298, opened 2026-03-26
Add discoveries api endpoint. This GET endpoint returns payment railes available by currency and country

matthappens:
### TL;DR

Added a new `/discoveries` endpoint to retrieve available payment rails for specific country and currency combinations.

### What changed?

- Added a new "Discoveries" tag and endpoint group for payment rail discovery functionality
- Implemented `GET /discoveries` endpoint that accepts `country` (ISO 3166-1 alpha-2) and `currency` (ISO 4217) query parameters
- The endpoint returns a list of payment rails with stable identifiers, names, display names, country, and currency information
- Each payment rail's `id` field serves as the stable identifier to use when creating external accounts
- Included comprehensive error handling for 400, 401, and 500 status codes
- Added example response showing Philippine banks (BDO Unibank and Bank of the Philippine Islands)

### How to test?

1. Make a GET request to `/discoveries?country=PH&currency=PHP` with proper authentication
2. Verify the response contains a `payment_rails` array with bank information
3. Test error cases by omitting required parameters or using invalid country/currency codes
4. Confirm the returned `id` values can be used as `bankName` when creating external accounts via `POST /customers/{id}/external-accounts`

### Why make this change?

This endpoint enables developers to programmatically discover which payment rails, banks, and providers are available for specific country-currency corridors supported by Grid's partners, eliminating the need to hardcode bank names and providing a dynamic way to populate payment options in applications.

### DIFF
--- a/openapi/paths/discoveries/discoveries.yaml
+++ b/openapi/paths/discoveries/discoveries.yaml
@@ -1,0 +1,93 @@
+get:
+  summary: List available receiving institution names
+  description: |
+    Retrieve available payment institution names for a given country and currency. Use this endpoint
+    to look up supported banks and payment providers for a specific corridor.
+    If no country and currency parameter are provided, all payment institutions will be returned
+
+    The `bankName` field in each result is the value to pass as `bankName` when
+    creating an external account via `POST /customers/external-accounts`.
+  operationId: getDiscoveries
+  tags:
+    - Discoveries
+  security:
+    - BasicAuth: []
+  parameters:
+    - name: country
+      in: query
+      description: ISO 3166-1 alpha-2 country code (e.g. PH)
+      required: false
+      schema:
+        type: string
+      example: PH
+    - name: currency
+      in: query
+      description: ISO 4217 currency code (e.g. PHP)
+      required: false
+      schema:
+        type: string
+      example: PHP
+  responses:
+    "200":
+      description: Successful operation
+      content:
+        application/json:
+          schema:
+            type: object
+            properties:
+              data:
+                type: array
+                description: List of payment rails matching the filter criteria
+                items:
+                  type: object
+                  required:
+                    - bankName
+                    - displayName
+                    - country
+                    - currency
+                  properties:
+                    bankName:
+                      type: string
+                      description: >
+                        Canonical name for this payment rail. Pass this value as
+                        `bankName` when creating an external account.
+                    displayName:
+                      type: string
+                      description: Human-friendly display name for this payment rail.
+                    country:
+                      type: string
+                      description: ISO 3166-1 alpha-2 country code.
+                    currency:
+                      type: string
+                      description: ISO 4217 currency code.
+          examples:
+            philippinesBanks:
+              summary: Payment rails for Philippines (PHP)
+              value:
+                data:
+                  - bankName: BDO Unibank
+                    displayName: BDO Unibank
+                    country: PH
+                    currency: PHP
+                  - bankName: BPI
+                    displayName: Bank of the Philippine Islands
+                    country: PH
+                    currency: PHP
+    "400":
+      description: Bad request - Invalid parameters
+      content:
+        application/json:
+          schema:
+            $ref: ../../components/schemas/errors/Error400.yaml
+    "401":
+      description: Unauthorized
+      content:
+        application/json:
+          schema:
+            $ref: ../../components/schemas/errors/Error401.yaml
+    "500":
+      description: Internal service error
+      content:
+        application/json:
+          schema:
+            $ref: ../../components/schemas/errors/Error500.yaml
--- a/mintlify/openapi.yaml
+++ b/mintlify/openapi.yaml
@@ -44,6 +44,8 @@ tags:
     description: Endpoints to programmatically manage API tokens
   - name: Exchange Rates
     description: Endpoints for retrieving cached foreign exchange rates. Rates are cached for approximately 5 minutes and include platform-specific fees.
+  - name: Discoveries
+    description: Endpoints for discovering available payment rails, banks, and providers for a given country and currency corridor.
 paths:
   /config:
     get:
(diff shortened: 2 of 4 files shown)


matthappens:
* **#298** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/298?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/298?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

openapi/paths/discoveries/discoveries.yaml
```diff
@@ -0,0 +1,103 @@
...
+                    - id
+                    - name
+                    - display_name
+                    - country
+                    - currency
+                  properties:
+                    id:
+                      type: string
+                      description: >
+                        Stable identifier for this payment rail. Opaque string —
```
matthappens: todo - this description gives too much away, stable id is probably sufficient

openapi/paths/discoveries/discoveries.yaml
```diff
@@ -0,0 +1,100 @@
...
+    to look up supported banks and payment providers for a specific corridor.
+
+    The `id` field in each result is the stable identifier to pass as `bankName` when
+    creating an external account via `POST /customers/{id}/external-accounts`.
+  operationId: getDiscoveries
+  tags:
+    - Discoveries
+  security:
+    - BasicAuth: []
+  parameters:
```
pengying, line 15: should this be a searchable list?  Eg instead of a drop down of however many banks, it's a searchable list and we return options starting with the search param they pass in.

openapi/paths/discoveries/discoveries.yaml
```diff
@@ -0,0 +1,100 @@
...
+      example: PHP
+  responses:
+    "200":
+      description: Successful operation
+      content:
+        application/json:
+          schema:
+            type: object
+            properties:
+              payment_rails:
```
pengying: This should follow our standard pagination scheme I think?  https://github.com/lightsparkdev/grid-api/tree/main/openapi#pagination

openapi/paths/discoveries/discoveries.yaml
```diff
@@ -0,0 +1,100 @@
...
+                items:
+                  type: object
+                  required:
+                    - id
+                    - name
+                    - display_name
+                    - country
+                    - currency
+                  properties:
+                    id:
```
pengying: what's teh difference between id, name and display name?  Should it only be id / display_name?

openapi/paths/discoveries/discoveries.yaml
```diff
@@ -0,0 +1,100 @@
...
+                        `bankName` when creating an external account.
+                    name:
+                      type: string
+                      description: Canonical name for this payment rail (e.g. "BDO Unibank").
+                    display_name:
+                      type: string
+                      description: >
+                        Human-friendly display name. Falls back to name if the provider
+                        has not set a distinct display name.
+                    country:
```
pengying, line 57: i forget the use case for including these.  Is it if a country supports multiple currencies?  or if a currency is across multiple countries?   
  
Should the country / currency url params both be required instead?

pengying: Just trying to think about how SoFi and others would build the flow.  I think the user would have already selected the country + currency

openapi/paths/discoveries/discoveries.yaml
matthappens: I'm open to this list - would you structure that as /discoveries providing a string parameter to narrow down the results?

openapi/paths/discoveries/discoveries.yaml
matthappens: short answer, yes should be id and display_name. long answer, the name field is a kludge because the external account integration api requires a name  rather than an id for bankName / provider etc.  I think the right thing to do would be to change the external account to require an id instead, then have discoveries return {id, display_name}

openapi/paths/discoveries/discoveries.yaml
matthappens: the immediate use case for us is currencies used in multiple countries, like XOF. we don't support this yet, but conceivably we could have multiple currencies per country, like EUR & DKK for denmark

matthappens:
@pengying trying to remember, we were 'no' on pagination for now?

pengying:
Yeah - if they're going to implement filtering on the client, I think pagination would slow down the UI.  And 100 * ~100-150 bytes is not terrible.

matthappens:
> Yeah - if they're going to implement filtering on the client, I think pagination would slow down the UI.  And 100 * ~100-150 bytes is not terrible.

​done, pagination removed

openapi/paths/discoveries/discoveries.yaml
shreyav: if country and currency are required on the request, and we limit to those, is there any benefit to returning them for every entry on the response?

openapi/paths/discoveries/discoveries.yaml
```diff
@@ -0,0 +1,92 @@
...
+          examples:
+            philippinesBanks:
+              summary: Payment rails for Philippines (PHP)
+              value:
+                data:
+                  - bankName: BDO Unibank
+                    displayName: BDO Unibank
+                    country: PH
+                    currency: PHP
+                  - bankName: Bank of the Philippine Islands
```
pengying: is the bankName / displayName for thie example reversed?

openapi/paths/discoveries/discoveries.yaml
pengying: I don't think there is.  But I was ok leaving them since I'm not sure exactly how  integrators will use it.   

Does it make sense to make country required?

shreyav: it seems kind of weird to me.. id rather leave it out and add if we see a need later

pengying: i don't feel too strongly either way

matthappens: we don't need to require it.  I think the present list of bank names is small enough we could send it back in one block. I assume this api will most often be called with a country_code parameter

openapi/paths/discoveries/discoveries.yaml
matthappens: yes! will fix

openapi/paths/discoveries/discoveries.yaml
matthappens: updated, these parameters are no longer required

openapi/paths/discoveries/discoveries.yaml
shreyav: \+1 on the list

pengying (approved):

openapi/paths/discoveries/discoveries.yaml
```diff
@@ -0,0 +1,92 @@
+get:
+  summary: List payment rails
```
shreyav: should this description be something different? bc theres no rails in the response

shreyav (approved):

openapi/paths/discoveries/discoveries.yaml
shreyav: discussed offline

matthappens:
### Merge activity

* **Mar 31, 8:57 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/lightsparkdev/grid-api/298).
* **Mar 31, 8:57 PM UTC**: @matthappens merged this pull request with [Graphite](https://app.graphite.com/github/pr/lightsparkdev/grid-api/298).

PR #333, opened 2026-04-10
docs: add MCP server to Building with AI page

pengying:
## Summary
- Adds an "MCP server" section to the Building with AI docs page
- Documents the `grid.lightspark.com/mcp` endpoint with its two tools (`search_grid_api_documentation`, `get_page_grid_api_documentation`)
- Includes a JSON config snippet for MCP-compatible clients
- Notes that the server supports documentation access for agentic workflows, not direct API interaction
- Placed after the agent skill section, before example prompts

## Test plan
- [ ] Verify the page renders correctly with `make mint`
- [ ] Confirm the anchor link to `#install-the-grid-api-agent-skill` works from the Info callout

### DIFF
--- a/mintlify/platform-overview/building-with-ai.mdx
+++ b/mintlify/platform-overview/building-with-ai.mdx
@@ -105,6 +105,35 @@ It'll prompt you for your API Token ID and Client Secret, validate them, and sav
 Start in the sandbox environment to experiment safely. The Skill is great at generating fake account data to help you test different flows.
 </Tip>
 
+## MCP server
+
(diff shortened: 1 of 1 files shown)


jklein24 (approved):
I still don't entirely see the benefit of using MCP over other mechanisms for this (the skill or just pointing at llms.txt), but I guess this doesn't hurt.

pengying:
I'm not sure either, just trying to fulfill all user preferences.  And the MCP server has 0 maintenance.  
  
It's after the skill which is preferred and we have more control over.

pengying:
I don't want some PM comparing between Grid and Modern Treasury and choosing Modern Treasury because they have a MCP server

PR #417, opened 2026-04-28
fix(grid-visualizer): emit documented minimum beneficiary fields

pengying:
The generated curl examples included `birthDate` and `nationality` on
every beneficiary (both are optional per docs) and omitted `address`
even for countries where docs say it's required. Aligns the output
with the minimum-required-fields table in the External Accounts docs:

- All countries: emit only beneficiaryType + fullName as the universal
  minimum (drop birthDate / nationality — both are optional, "providing
  additional information... can reduce false positive compliance checks").
- US (USD), Europe (EUR), UK (GBP): include a country-appropriate
  beneficiary.address (line1, city, postalCode, country, optional state).
- Mexico (MXN), Brazil (BRL), Philippines (PHP): explicit
  beneficiaryAddressRequired: false — docs call out that address is not
  required for these.
- Other account types: left undefined so the generated payload still
  emits the universal minimum, since the docs only document the
  minimum-fields table for the six countries above.

Adds BeneficiaryAddressExample type + two optional fields on
AccountTypeSpec (beneficiaryAddressRequired, beneficiaryAddressExample).
The code-generator's buildAccountInfoBody is updated to include the
address block conditionally.

Ref: https://grid.lightspark.com/payouts-and-b2b/depositing-funds/external-accounts#individual-beneficiaries

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/components/grid-visualizer/src/data/account-types.ts
+++ b/components/grid-visualizer/src/data/account-types.ts
@@ -4,10 +4,26 @@ export interface AccountFieldSpec {
   description?: string;
 }
 
+export interface BeneficiaryAddressExample {
+  line1: string;
+  city: string;
+  state?: string;
+  postalCode: string;
+  country: string;
+}
+
 export interface AccountTypeSpec {
   accountType: string;
   fields: AccountFieldSpec[];
   beneficiaryRequired: boolean;
+  // Whether the beneficiary needs a postal address. Per docs
+  // (https://grid.lightspark.com/payouts-and-b2b/depositing-funds/external-accounts):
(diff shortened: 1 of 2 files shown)


pengying:
* **#417** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/417?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/417?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #469, opened 2026-05-13
add internal account status

wuvictor-95:


wuvictor-95:
* **#469** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/469?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/469?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jklein24 (approved):
Would be cool to also update the mintlify docs if any relevant examples need to change.

openapi/components/schemas/customers/InternalAccountStatus.yaml
```diff
@@ -0,0 +1,19 @@
...
+  - OPEN
+  - CLOSED
+  - FROZEN
+description: >-
+  Status of a Grid internal account.
+
+
+  - `PENDING`: Account is pending activation.
+
+  - `OPEN`: Account is open and active.
```
pengying: can you add more information about the impact of each account state on transactions?  Eg can a pending account receive payments?  What about recurring debits on frozen accounts?

eg my account is frozen but my monthly phone bill auto debits from the account.  does that succeed or fail?

pengying: maybe ACTIVE is a better?

openapi/components/schemas/customers/InternalAccountStatus.yaml
```diff
@@ -0,0 +1,19 @@
...
+
```
pengying, line 17: Can you add a state diagram with markdown?

Can an account be closed with a non zero balance?  Or is there a PENDING_CLOSED where debits can occur but no credits before the account is closed?  
  
What triggers an account to go from PENDING > ACTIVE?

openapi/components/schemas/customers/InternalAccountStatus.yaml
```diff
@@ -0,0 +1,19 @@
...
+  - `CLOSED`: Account is closed.
```
pengying: Can integrators edit these?  For example if one of their customers requests their account to be closed is that a patch here?

openapi/components/schemas/customers/InternalAccountStatus.yaml
```diff
@@ -0,0 +1,19 @@
+title: Internal Account Status
```
shreyav, line 1: should we also send webhooks on status changes?

openapi/components/schemas/customers/InternalAccountStatus.yaml
wuvictor-95: No, these are updated from our banking provider.

openapi/components/schemas/customers/InternalAccountStatus.yaml
wuvictor-95: I had similar questions. Erebor said an account can be frozen or closed entirely for various compliance reasons. Frozen can be recovered, closed cannot. I would imagine the account could not receive or send money in these states but I can double check. 

PENDING means the account is being provisioned and undergoing review by the banking partner. Once approved, it moves to OPEN

openapi/components/schemas/customers/InternalAccountStatus.yaml
pengying: but what if an end customer says hey I want to close my account

wuvictor-95: In that case, the only viable option for a user-initiated patch on the internal account status is `CLOSED` and nothing else. Does that sound right to you?

openapi/components/schemas/customers/InternalAccountStatus.yaml
wuvictor-95: Added

openapi/components/schemas/customers/InternalAccountStatus.yaml
pengying: hey someone stole my card

wuvictor-95: I'm asking if that validation logic sounds right to you. If so, I can add it to the patch handler so integrators can expose an option to close their account.

pengying: Yeah that patch to close sounds right?  I think it ties back into can they close an account with a non zero balance.  But patch feels like the right mechanism

wuvictor-95: I think it makes sense to allow

openapi/components/schemas/customers/InternalAccountStatus.yaml
wuvictor-95: Changed to ACTIVE

wuvictor-95: Added some more context

openapi/components/schemas/customers/InternalAccountStatus.yaml
```diff
@@ -0,0 +1,25 @@
...
+  Status of a Grid internal account. The status determines whether the account
+  can send or receive payments.
+
+
+  - `PENDING`: The account is under review and is being provisioned. The account
+  cannot send or receive payments until provisioning completes.
+
+  - `ACTIVE`: The account is ready to send and receive payments.
+
+  - `CLOSED`: The account cannot send or receive payments. A customer can
```
pengying, line 18: closed is terminal?  No undo right?

wuvictor-95: Correct. Some new information here: https://docs.erebor.bank/accounts/deposit-accounts#account-lifecycle

I'm just seeing this since Erebor did not have `guides` before. But looks like we can follow the same logic of closing it down when balance is 0. I can add some more validation.

pengying: cool.  what's the password to the docs?

jklein24: @wuvictor-95 please don't respond to that question here in this public repo :-)

pengying (approved):

PR #540, opened 2026-06-02
Add receipt delivery confirmation endpoint

bsiaotickchong:
## Summary
- add POST /transactions/{transactionId}/confirm to confirm transaction receipt delivery
- allow an optional receiptDeliveryConfirmedAt timestamp, with server-time default handled by the implementation
- expose receiptDeliveryConfirmedAt on transaction responses
- return the updated TransactionOneOf from the confirm endpoint

## Test plan
- npm run lint


bsiaotickchong:
@greptileai

pengying (approved):

PR #552, opened 2026-06-05
Add CnyExternalAccountCreateInfo + register in CreateInfo OneOf

mohamedwane:
## Summary

- Adds the missing `CnyExternalAccountCreateInfo.yaml` (hand-authored, mirrors the existing `BdtExternalAccountCreateInfo.yaml` template) and registers it in `ExternalAccountCreateInfoOneOf.yaml`.
- The auto-sync workflow's `create_external_accounts.py` only generates the read-side `{Currency}ExternalAccountInfo.yaml`; the create-side `{Currency}ExternalAccountCreateInfo.yaml` is hand-authored per corridor and was missed when CNY landed via #551. Without the create variant in the OneOf, the discriminator can't dispatch `accountType: CNY_ACCOUNT` on `POST /customers/external-accounts` and the request fails with `"did not match any expected shape"`.

## Test plan

- [x] `npm run build:openapi` produces a bundle that contains `CnyExternalAccountCreateInfo` (line 15259 in the regenerated `openapi.yaml`).
- [ ] After merge, regen the downstream sparkcore mirror via `update_schema.sh` and confirm `grid_api/models/cny_external_account_create_info.py` exists.
- [ ] Once mirror regen lands in sparkcore, EGP-style CNY create payload accepts.

### DIFF
--- a/mintlify/openapi.yaml
+++ b/mintlify/openapi.yaml
@@ -14616,6 +14616,26 @@ components:
                 mapping:
                   INDIVIDUAL: '#/components/schemas/CadBeneficiary'
                   BUSINESS: '#/components/schemas/BusinessBeneficiary'
+    CnyExternalAccountCreateInfo:
+      title: CNY Account
+      allOf:
+        - $ref: '#/components/schemas/BaseExternalAccountInfo'
+        - $ref: '#/components/schemas/CnyAccountInfoBase'
+        - type: object
+          required:
+            - beneficiary
+          properties:
+            beneficiary:
+              oneOf:
+                - title: Individual Beneficiary
+                  $ref: '#/components/schemas/CnyBeneficiary'
+                - title: Business Beneficiary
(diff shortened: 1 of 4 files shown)


mohamedwane:
Good catch — fixed in 4451e8b. Added `CNY_ACCOUNT: ./CnyExternalAccountCreateInfo.yaml` to the `discriminator.mapping` block (between `CAD_ACCOUNT` and `COP_ACCOUNT`) and re-ran `npm run build:openapi`. The bundled `openapi.yaml` now carries `CNY_ACCOUNT: '#/components/schemas/CnyExternalAccountCreateInfo'` in the mapping. Also verified the read-side `ExternalAccountInfoOneOf.yaml` already had the matching `CNY_ACCOUNT` entry (auto-synced via #551), so this PR closes the gap on the create side.

JasonCWang (approved):

PR #558, opened 2026-06-08
feat(sca): add Strong Customer Authentication surface

jklein24:
## Summary

The **per-transaction SCA authorization surface** — everything needed to gate and authorize an EU (Striga) money-movement debit with an SMS OTP or passkey. This is the compliance-critical core of the Striga SCA work and the **base of the grid-api SCA stack**; the management surface (#600) stacks on top, and the sparkcore backend lands separately. Designed to stay **invisible to every non-SCA (non-EU) caller**.

## Covered by this PR

**Shared SCA vocabulary** (defined here, reused by the whole stack):
- `ScaFactor` (`SMS_OTP` / `TOTP` / `PASSKEY`)
- `ScaChallenge` (`id`, `expiresAt`, `factor`, `availableFactors`, `passkeyAssertionOptions`, `passkeyAllowedOrigins`)
- `ScaAuthorization` (exactly one of `code` / `passkeyAssertion`; `origin` required with a passkey assertion)

**Per-transaction authorization:**
- `PENDING_AUTHORIZATION` status + readOnly `scaChallenge` on `Quote` and `Transaction` (incl. the `OutgoingTransactionStatus` override) — omitted entirely for non-SCA providers.
- Resource-scoped authorize endpoints: `POST /quotes/{id}/authorize` → `Quote`, `POST /transactions/{id}/authorize` → `Transaction` (both `409` for non-SCA providers).
- SMS code re-send: `POST /quotes/{id}/authorize/resend`, `POST /transactions/{id}/authorize/resend`.
- Inline one-shot proof: optional `scaAuthorization` on `execute` / `transfer-out` (satisfy the challenge in the same call).
- Optional `scaFactor` on `execute` / `transfer-out` to pick `SMS_OTP` (default) or `PASSKEY` for the per-transaction challenge.

## NOT in this PR — see #600 (stacked on top)

The SCA **management** surface (customer-level setup/recovery, not per-payment authorization) lives in #600 and reuses the schemas defined here:
- Factor **enrollment** — TOTP + passkey register / list / delete
- **Login / 180-day session** — `sca/login/start|complete`
- **Beneficiary trust** whitelisting — `external-accounts/{id}/trust`, `.../untrust/confirm`
- **2FA reset / recovery** — liveness-gated
- Security-event recording — `sca/record-event`

Also out of scope: the **sparkcore backend** (separate stack) and any client-side WebAuthn ceremony logic (Grid relays opaque passkey blobs; it performs no crypto).

## Known caveat

[...]

PR #600, opened 2026-06-17
feat(sca): SCA management surface — enrollment, login/session, beneficiary trust, 2FA reset

jklein24:
## Summary

The SCA **management** surface for EU (Striga) customers — the customer-level setup, session, and recovery flows that sit **alongside** per-payment authorization (not inside it). This single PR collapses the original design's **PR3–PR7** (enrollment, passkey, login/session, beneficiary trust, 2FA reset). Stacks on **#558**, which defines the per-transaction authorization core and the shared schemas this PR reuses. All endpoints are **EU-only — `409` for providers that don't require SCA** (the invisibility guarantee) and mirror the sparkcore handlers' exact routes + JSON shapes. Surfaced after an audit of #558 found these existed only as sparkcore temp types, not in the spec.

## Covered by this PR

- **Factor enrollment** (design PR3/PR4): `POST /customers/{id}/sca/factors/totp` (+`/confirm`), `.../passkey` (+`/confirm`), `GET .../sca/factors`, `DELETE .../passkey/{credentialId}`.
- **Login / 180-day session** (PR5): `POST .../sca/login/start` (+`/complete`); `POST .../sca/record-event`.
- **Beneficiary trust** (PR6): `POST .../external-accounts/{extId}/trust` (issues the challenge + `whitelistedId`), `.../trust/confirm`, `.../untrust/confirm` — the whitelisting exemption that lets recurring payees skip per-transaction SCA.
- **2FA reset / recovery** (PR7): `POST .../sca/factors/reset` (201) → `GET .../reset/{resetId}` (poll to `LIVENESS_PASSED`) → `POST .../reset/{resetId}/complete`.

Reuses `ScaFactor` / `ScaChallenge` / `ScaAuthorization` from #558; confirm/login bodies carry the `code | passkeyAssertion + origin` proof, and beneficiary confirm threads `whitelistedId` back so confirm never re-whitelists (no duplicate SMS). `TwoFactorResetStatus` / `ScaLoginComplete` document their provider-passthrough terminal values (`LIVENESS_PASSED` / `SUCCESS`).

## NOT in this PR

- **Per-transaction authorization** — the challenge on a debit, the `authorize` / `resend` endpoints, and inline `scaAuthorization` / `scaFactor` — that's **#558**.
- The **sparkcore backend** for these endpoints (separate stack), and client-side WebAuthn ceremony logic (Grid relays opaque passkey blobs; it performs no crypto).

## Known gaps / follow-ups (tracked in the design doc addendum)

- **Untrust** has `confirm` but no `start`, and there is **no trust-list read** — a caller who loses `whitelistedId` (or needs the untrust OTP re-sent) is stuck (addendum **F4**). The trusted-beneficiary *exemption benefit* (a trusted payee actually skipping per-tx SCA) is also not yet realized in the backend.
- Reset `status` documents `LIVENESS_PASSED` / `PENDING` but no terminal **failure** value — infinite-poll risk (**F10**).
- Distinct SCA `409` codes (`SCA_NOT_REQUIRED` vs `NO_PENDING_SCA_CHALLENGE`) are not yet modeled (**F9**).

## Validation

`npm run lint:openapi` (redocly + spectral, fail-on-error) passes clean; additions are additive / non-breaking. (The OpenAPI build / breaking-changes / lint CI checks run on the base=`main` PR #558; they do not run here because this PR targets `feat/striga-sca`.)

## Notes

- Field shapes mirror the sparkcore implementation (the source of truth until the SDK regenerates). A few Striga response bodies (login-complete, reset-status/complete, passkey list/delete) are typed from the sparkcore `to_dict()` since Striga's docs only cover the start side — flagged for sandbox verification.

## Stacking

`main` → #558 (per-transaction) → **#600 (this PR)**. Merges after #558; depends on the schemas #558 introduces.


jklein24:
@greptile review

openapi/components/schemas/sca/TwoFactorResetStatus.yaml
```diff
@@ -0,0 +1,11 @@
...
+description: >-
+  The provider-reported status of an in-progress 2FA reset, polled until it
+  reaches the provider's liveness-passed value.
+required:
+  - status
+properties:
+  status:
+    type: string
+    description: The provider-reported status of the reset.
+    example: PENDING
```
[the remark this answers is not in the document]
jklein24: **Applied.** Enriched the `status` description to name the terminal sentinel (`LIVENESS_PASSED`) and explain it's a verbatim provider passthrough that consumers should poll against, treating any non-terminal value as "not ready yet." Kept it a free-form string rather than a locked `enum` because the implementation deliberately passes Striga's value through unnormalized — an enum would reject valid future provider states.

_Responded via Claude Code (pr-review-respond)._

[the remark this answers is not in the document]
jklein24: **Applied.** Added the `LIVENESS_PASSED` terminal value (and the `SUCCESS` value on `ScaLoginComplete.status`) to the field descriptions so consumers have an in-spec stop condition. Used a description note rather than the suggested `enum`/strict form since the status is a verbatim provider passthrough.

_Responded via Claude Code (pr-review-respond)._

openapi/components/schemas/sca/TotpEnrollmentConfirmRequest.yaml
```diff
@@ -0,0 +1,20 @@
...
+  secret:
+    type: string
+    description: >-
+      The shared secret returned as `secret` by the TOTP enrollment start,
+      threaded back to bind the confirmation to that enrollment.
+  code:
+    type: string
+    description: >-
+      The current time-based one-time code from the customer's authenticator
+      app. In sandbox, the code is always `123456`.
```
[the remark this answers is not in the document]
jklein24: **Intentional — not changing.** The `secret`/`code` round-trip is the actual confirm contract (mirrored by the sparkcore impl and its integration test: `/sca/factors/totp/confirm` body is `{secret, code}`). The enrollment is stateless — Grid persists no pending-enrollment record — so the client threads the `secret` back to bind the confirmation to the originating start, analogous to (but distinct from) the challenge-id pattern. Reworking this into a server-side opaque token would diverge the spec from the provider contract it documents. The field description already states the binding purpose.

_Responded via Claude Code (pr-review-respond)._

jklein24:
@greptile review

jklein24:
@greptile review

jklein24:
* **#683** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/683?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#600** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/600?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/600?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#558** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/558?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

jklein24:
@faraday review this

openapi/components/schemas/sca/BeneficiaryTrustConfirmRequest.yaml
```diff
@@ -0,0 +1,47 @@
+type: object
+description: >-
+  Confirms trusting or untrusting a beneficiary by submitting the SCA proof.
+  Carries the same proof fields as an `ScaAuthorization` (`code` for `SMS_OTP` /
+  `TOTP`, or `passkeyAssertion` + `origin` for `PASSKEY`), plus the
+  `whitelistedId` returned by the trust start and, when the start issued one,
+  the `challengeId`.
+required:
```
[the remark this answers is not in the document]
jklein24: **Applied.** Added `anyOf: [{required: [code]}, {required: [passkeyAssertion, origin]}]` to `BeneficiaryTrustConfirmRequest` and `ScaLoginCompleteRequest`, so a proof-less body now fails schema validation instead of only at runtime. (Suppressed the redocly `no-required-schema-properties-undefined` rule for these via `.redocly.lint-ignore.yaml`, following the existing precedent for `QuoteSource` / `Quote.destination` — the construct is valid JSON Schema.)

_Review response generated using `/pr-review-respond` Claude command._

openapi/components/schemas/sca/RecordSecurityEventRequest.yaml
```diff
@@ -0,0 +1,12 @@
+type: object
+description: >-
+  Records a client-side security-relevant event for the customer with the
+  underlying provider's risk engine (e.g. a sign-in, a sensitive view). Used to
+  feed the provider's adaptive-authentication signals.
+required:
+  - eventType
+properties:
+  eventType:
```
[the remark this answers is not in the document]
jklein24: **Applied.** Documented that `eventType` is a provider-defined closed vocabulary forwarded verbatim to the provider's risk engine (which rejects values outside its set), not free-form caller input. Kept it a string rather than a hard enum so new provider-defined events do not force a spec change.

_Review response generated using `/pr-review-respond` Claude command._

openapi/components/schemas/sca/BeneficiaryTrustConfirmRequest.yaml
```diff
@@ -0,0 +1,47 @@
...
+  - whitelistedId
+properties:
+  whitelistedId:
```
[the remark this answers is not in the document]
jklein24: **Applied.** Documented that `whitelistedId` is scoped to *this account's own* trust-start — the server rejects one issued for a different external account or from an expired/stale ceremony, so callers must not cache or reuse it across accounts.

_Review response generated using `/pr-review-respond` Claude command._

openapi/components/schemas/sca/TwoFactorResetStart.yaml
```diff
@@ -0,0 +1,36 @@
...
+  ceremony together (status and complete reference it). `sumsubAccessToken` and
+  `verificationLink` are omitted when the provider does not return them.
+required:
+  - resetId
+properties:
+  resetId:
+    type: string
+    description: >-
+      Identifier for this reset; pass it to the status and complete endpoints.
+  sumsubAccessToken:
```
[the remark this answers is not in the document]
jklein24: **Applied.** Renamed `sumsubAccessToken` → `livenessAccessToken` (provider-neutral) here **and** in the sparkcore `TwoFactorResetStartResponse` in lockstep, before SDK generation makes it a breaking change. The Striga client adapter keeps the `sumsub` name internally since it mirrors Striga's actual wire field; status values remain a documented provider passthrough (a normalized status enum is noted as a separate follow-up).

_Review response generated using `/pr-review-respond` Claude command._

openapi/components/schemas/sca/RecordSecurityEventRequest.yaml
```diff
@@ -0,0 +1,18 @@
...
+    type: string
+    description: >-
+      A provider-defined security-event identifier drawn from the provider's
```
jklein24: I think we may need to be more explicit and intentional here. We should just explicitly use striga's values if this endpoint is just meant to follow https://docs.striga.com/reference/recorduserevent.md. Just mirror it including the description rather than trying to be abstract about the underlying provider that's totally abstracted away from the caller.

openapi/components/schemas/sca/BeneficiaryTrustStart.yaml
```diff
@@ -0,0 +1,21 @@
+type: object
+description: >-
+  The provider handle plus the SCA challenge a caller authorizes to finish
+  trusting (or untrusting) a beneficiary. `whitelistedId` is the provider's
```
jklein24: Here and in other related endpoints, it's kind of awkward that we're exposing striga guts here with the provider's "whitelistedId" being a separate ID from the actual external account ID. Is there any way we can easily abstract this away from the API surface if we just store the external account ID mapping to the underlying striga beneficiary ID internally in our backend?

openapi/components/schemas/sca/RecordSecurityEventRequest.yaml
jklein24: **Applied.** `eventType` is now an explicit `enum` mirroring the two real values, with each one's effect documented (framed as Grid's, no underlying-provider mention):
- `RESET_PASSWORD_COMPLETED` — revokes every active SCA session and clears the failed-login counter.
- `FAILED_LOGIN_ATTEMPT` — increments the failed-login counter and escalates lockout (5→15m, 6→30m, 7→1h, 8→24h, 9+→suspension).

Dropped the vague "closed vocabulary" free-string and the wrong `LOGIN` example. Note: this tightens the contract to those two values — the backend handler should validate against them.

_Review response generated using `/pr-review-respond` Claude command._

openapi/components/schemas/sca/BeneficiaryTrustStart.yaml
jklein24: **Applied (spec).** Removed `whitelistedId` from the trust surface entirely. `BeneficiaryTrustStart` now returns just the `scaChallenge`, and `BeneficiaryTrustConfirmRequest` no longer carries it — trust / confirm / untrust all identify the beneficiary by the `externalAccountId` already in the path, so the caller never sees a separate handle.

This needs a **backend follow-up**: Grid must persist the `externalAccountId → whitelist handle` mapping and resolve it internally on confirm/untrust (same contract-ahead-of-backend pattern as the rest of the stack). Flagging so the sparkcore side is wired before this ships.

_Review response generated using `/pr-review-respond` Claude command._

pengying:
Can we reuse some of the patterns we used for Embedded wallet auth?    
  
Creation  
/auth/credentials   
https://docs.lightspark.com/api-reference/embedded-wallet-auth/create-an-authentication-credential  
  
Delete  
https://docs.lightspark.com/api-reference/embedded-wallet-auth/revoke-an-authentication-credential  
  
Challenge  
https://docs.lightspark.com/api-reference/embedded-wallet-auth/re-issue-an-authentication-credential-challenge  
  
Verify  
https://docs.lightspark.com/api-reference/embedded-wallet-auth/verify-an-authentication-credential

openapi/paths/customers/customers_external_accounts_{externalAccountId}_trust.yaml
```diff
@@ -0,0 +1,65 @@
+parameters:
```
pengying, line 1: Maybe this could live under Sca ? /sca/trusted-external-accounts/challenge with  
  
{  
  customerId: id,  
  externalAccountId: id  
}  
  
Then /sca/trusted-external-accounts/verify

openapi/openapi.yaml
```diff
@@ -153,6 +153,38 @@ paths:
...
+  /customers/{customerId}/sca/factors/reset:
+    $ref: paths/customers/customers_{customerId}_sca_factors_reset.yaml
+  /customers/{customerId}/sca/factors/reset/{resetId}:
+    $ref: paths/customers/customers_{customerId}_sca_factors_reset_{resetId}.yaml
+  /customers/{customerId}/sca/factors/reset/{resetId}/complete:
+    $ref: paths/customers/customers_{customerId}_sca_factors_reset_{resetId}_complete.yaml
+  /customers/{customerId}/external-accounts/{externalAccountId}/trust:
+    $ref: paths/customers/customers_{customerId}_external-accounts_{externalAccountId}_trust.yaml
+  /customers/{customerId}/external-accounts/{externalAccountId}/trust/confirm:
+    $ref: paths/customers/customers_{customerId}_external-accounts_{externalAccountId}_trust_confirm.yaml
```
pengying: does untrust need a verify?

pengying: could it be like deleted a trusted account resource?

pengying: or maybe i SCA when adding the external account?

jklein24: I think we already have that below?

pengying: Sorry 3 different thoughts  
  
1\. If we model trusted external accounts as a resource eg /sca/trusted-external-accounts/{id}   
Could untrusting it be a deletion.  

1. Does the untrust process also need a SCA or can it be a simple delete it?
2. Does it make sense to add SCA to external account creation directly?  so it becomes create external account + verify sca.  Or i guess this SCA is not always required?  let me finish reading the SCA thread.

jklein24: It does require an SCA to untrust - https://docs.striga.com/reference/confirmbeneficiaryuntrust

openapi/paths/sca/sca_factors_reset_{resetId}_complete.yaml
prashb94, line 1: For reset 2FA completion, if it's SMS, this will take in a new mobile number

openapi/components/schemas/sca/TwoFactorResetStatus.yaml
prashb94, line 1: Maybe worth enumerating our REJECTED/EXPIRED as terminal so clients don't poll a failed status? our status response also carries factor, enrollmentStatus, expiresAt, completedAt, which are worth surfacing. Also: we rate limit reset initiation (5 per 24h), might be worth adding a 429 for that

openapi/paths/customers/customers_external_accounts_{externalAccountId}_trust.yaml
```diff
@@ -0,0 +1,58 @@
...
+    required: true
+    schema:
+      type: string
+post:
+  summary: Start trusting a beneficiary
+  description: |
+    Begin trusting (whitelisting) an external account so future sends to it can
+    skip the per-transaction SCA ceremony. Returns the `scaChallenge` to satisfy
+    (when one is issued). Complete with
+    `POST /customers/external-accounts/{externalAccountId}/trust/confirm`.
```
prashb94, line 14: @ronniebhatt please correct me if I'm wrong here but a trusted eur/usdc payout will still get a challenge and what's skipped is the dynamic linking enforcement no?   
  
Also only USDC addresses can be "trusted" right?

openapi/components/schemas/sca/RecordSecurityEventRequest.yaml
prashb94, line 1: Looks like the lockout state is invisible? HTTP 423 (code 31105). We return {event, suspended?, lockedUntil?, failedAttempts?} so might be worth surfacing lockedUntil to tell the user to try again in XX

openapi/paths/sca/sca_factors_passkey.yaml
```diff
@@ -0,0 +1,59 @@
...
+    required: true
+    schema:
+      type: string
+    example: Customer:019542f5-b3e7-1d02-0000-000000000001
+post:
+  summary: Start passkey factor enrollment
+  description: |
+    Begin enrolling a WebAuthn passkey factor for the customer. Returns opaque
+    WebAuthn registration `options`; pass them to the device's WebAuthn API and
+    submit the resulting credential via
```
prashb94: Please remember our current implementation is one passkey per user (We return error 100002 when a user tries to enrol a new passkey)

openapi/paths/customers/customers_external_accounts_{externalAccountId}_trust.yaml
jklein24: Moved a bunch of stuff.

openapi/paths/sca/sca_factors_reset_{resetId}_complete.yaml
jklein24: Good call. Confirmed it against the docs and added it — `complete` now takes a `mobile` object (`countryCode` + `number`), required for SMS resets. Docs updated to match.

openapi/components/schemas/sca/TwoFactorResetStatus.yaml
jklein24: Yep, all in. `status` is an enum now, with `COMPLETED`/`REJECTED`/`EXPIRED` marked terminal so clients stop polling a dead reset, and it surfaces `factor`, `enrollmentStatus`, `expiresAt`, `completedAt`. Added a `429` on reset initiate for the 5/24h limit too. Pulled the exact `status` and `enrollmentStatus` value sets from your docs.

openapi/paths/customers/customers_external_accounts_{externalAccountId}_trust.yaml
jklein24: Yeah, fixed the USDC part — docs now say only USDC addresses can be trusted (dropped the old "IBAN or crypto address" framing). On the challenge-vs-dynamic-linking piece I think you're right, but I'd rather have @ronniebhatt confirm the exact semantics before I reword it, since it's his flow. Leaving that half open for him.

openapi/components/schemas/sca/RecordSecurityEventRequest.yaml
jklein24: Good catch, the lockout state was totally invisible. Fixed — record-event returns `{ suspended, lockedUntil, failedAttempts }` now (200 instead of 204), plus a `423` (`ACCOUNT_LOCKED`) that surfaces `lockedUntil`/`failedAttempts` so integrators can tell the user when to retry. Added the same `423` to login-complete since a locked user hits it there too.

openapi/paths/sca/sca_factors_passkey.yaml
jklein24: Noted and documented. Enrolling a passkey now returns a `409` (`PASSKEY_ALREADY_ENROLLED`) on a second one, and the docs call out the one-per-user limit plus the fact that the passkey side of the factor list is 0-or-1.

openapi/components/schemas/sca/ScaFactorView.yaml
```diff
@@ -0,0 +1,21 @@
...
+required:
+  - factor
+properties:
+  factor:
+    $ref: ./ScaFactor.yaml
+    description: The kind of enrolled factor.
+  credentialId:
+    type:
+      - string
+      - 'null'
```
pengying: do we need explicit nulls here?  I saw this in another PR but always thought that the field would be excluded vs explicitly null'd out

openapi/paths/sca/sca_factors.yaml
```diff
@@ -0,0 +1,114 @@
...
+          schema:
+            $ref: ../../components/schemas/errors/Error500.yaml
+post:
+  summary: Start SCA factor enrollment
+  description: |
+    Begin enrolling an SCA factor for the customer. The request body's `type`
+    selects the factor — `TOTP` or `PASSKEY`. `SMS_OTP` is not enrollable (it uses
+    the customer's verified phone). Returns the factor-specific material needed to
+    finish via `POST /sca/factors/confirm`.
+
```
pengying, line 65: Is enrolling the concept of trusting a user device / auth method.  What triggers the sms otp?

pengying (commented):
LGTM 2 general questions  
1\. how does someone add sms otp  
2\. do we need -null as a type if the field is optional or are we explicitly setting it to null.

openapi/paths/sca/sca_factors.yaml
jklein24: Yes, it is. With striga, sms is enrolled by default since creating a customer requires sms and email verification via the contact verification API flows. I could definitely document this a bit better here, though. Striga does that at the top of this page: https://docs.striga.com/reference/enrollment. I'll clarify the SMS_OTP line here to indicate that all customers need a verified phone at customer creation in SCA-regulated regions. See the "verifications" routes.

pengying: so then when the customer is created, they provide the phone number as part of customer creation. we auto send a sms otp?

What if they change phone numbers? is that a patch to customer.phone_number that triggers a sms otp?  or do we make it explicit here?

openapi/components/schemas/sca/ScaFactorView.yaml
jklein24: Yeah, good call — dropped them. These are genuinely *omitted* for `TOTP`/`SMS_OTP`, not sent as `null`, so plain-optional is the right contract and the explicit null was just wrong here.

fwiw the rule I'm going by: nullable and optional are separate axes. Leave a field out of `required` when the key can be absent; only add `'null'` to the type when the server actually puts `"field": null` on the wire. A few of the newer sca/agents schemas started reaching for explicit null where the field is really just omitted — that's the inconsistency you're seeing. Fixing those as I hit them.

pengying: wait who dis?

shreyav (approved):

pengying (approved):

PR #631, opened 2026-06-29
docs(openapi): correct exchangeRate example to match documented basis (PROD-54)

AaryamanBhute:
## Summary

The `exchangeRate` field on the `/exchange-rates` `ExchangeRate` schema is documented as **"Number of sending currency units per receiving currency unit"**, but the examples were showing the inverse (destination-per-source) direction — contradicting both the field description and the live API:

- `ExchangeRate.yaml` schema example: USD→INR `82.50`
- `allRatesFromUSD` response example: USD→INR `82.50`, USD→EUR `0.925`

The live API returns `~0.01` for USD→INR, confirming the field is sending-units-per-receiving-unit. This is a docs-only example fix — the API itself is correct (confirmed by the ticket owner on PROD-54).

## Changes

Corrected the examples to the documented basis, keeping each response example internally consistent with its `sendingAmount`/`receivingAmount`:

| Corridor | Before | After | Basis check |
|----------|--------|-------|-------------|
| USD→INR `exchangeRate` | `82.50` | `0.012121` | `1/82.50`; $100 → ₹8,250 (receivingAmount 825000) ✓ |
| USD→EUR `exchangeRate` | `0.925` | `1.081081` | `1/0.925`; $100 → €92.50 |
| USD→EUR `receivingAmount` | `18500` | `9250` | €92.50 at the corrected rate ✓ |

Files:
- `openapi/components/schemas/exchange_rates/ExchangeRate.yaml`
- `openapi/paths/exchange-rates/exchange_rates.yaml`
- `openapi.yaml` + `mintlify/openapi.yaml` (regenerated via `make build`)

## Scope note

PROD-54 was split out of PROD-50 (the broader exchange-rate basis cleanup). The many `exchangeRate` examples in the narrative Mintlify docs (`0.92`, `17.25`, etc.) share the same directional inconsistency but are tracked under PROD-50 and intentionally **out of scope** here. The `Quote.yaml` `exchangeRate` field shares the same description but has no example, so nothing to fix there.

## Test plan
- `make build` — rebundles cleanly
- `make lint-openapi` — 0 errors (pre-existing warnings/infos on unrelated schemas only)
- Verified each corrected example is internally consistent with its `sendingAmount`/`receivingAmount`

Fixes PROD-54

Slack thread: https://lightsparkgroup.slack.com/archives/D0AHP8H9E13/p1782750571018079

---
[...]

PR #640, opened 2026-07-02
docs(mintlify): correct exchangeRate example basis (PROD-50)

Ethan-Minghao:
## Summary

#631 (PROD-54) fixed the exchangeRate example on the standalone
`ExchangeRate` schema, and noted the narrative Mintlify examples
(`0.92`, `17.25`, etc.) share the same directional inconsistency but
were intentionally out of scope, tracked separately under PROD-50.
This PR covers that remaining scope.

All three examples had `exchangeRate` in the inverse (destination-per-
source) direction, contradicting the field description ("Number of
sending currency units per receiving currency unit") and the live API
— verified against a live sandbox `/quotes` and `/exchange-rates` call,
both consistent with `(totalSendingAmount − feesIncluded) / exchangeRate
= totalReceivingAmount`.

Also includes an unrelated fix: `gridApiVariableFeeAmount` example (30)
didn't match `gridApiVariableFeeRate` (0.3%) applied to the stated
$1,000 transaction — should be 300.

## Changes

| Corridor | File | Before | After | Basis check |
|---|---|---|---|---|
| USD→BTC | quote-system.mdx | rate `0.00000833`, receiving `829167` | rate `120048.01920768`, receiving `828835` | `(1000−5) / 120048.01920768 = 0.00828835 BTC` |
| USD→EUR | quote-system.mdx | rate `0.92` | rate `1.08695652` | `995 / 1.08695652 = 915.40` (receiving amount unchanged) |
| USD→MXN | currencies-and-rails.mdx | rate `17.25`, receiving `1725000` | rate `0.05797101`, receiving `1716375` | `(1000−5) / 0.05797101 = 17163.75` |
| gridApiVariableFeeAmount | quote-system.mdx | `30` | `300` | `$1,000 × 0.3% = $3.00 = 300 cents` |

Prose captions updated to match in all three currency examples.

## Known limitations (not addressed in this PR)

- `rateDetails` components (`counterpartyFixedFee` + `gridApiFixedFee`
  + `gridApiVariableFeeAmount`) sum to 320, not the `feesIncluded: 500`
  shown at the top of the same example. Left as-is — insufficient
  information to determine the correct total without knowing whether
  additional fee components exist.
- `counterpartyMultiplier` breakdown text ("1.08 = 1 mSAT = 1.08 cents
  EUR") implies an implausible BTC/EUR price either with or without
  "cents" — left unchanged since neither reading is verifiable.
- `payouts-and-b2b/quickstart.mdx` has the same directional issue in
  three more places (plus an explicit "17.25 MXN per USD" label), but
  its example amounts don't cleanly round-trip under either direction
  (possibly a different underlying rate was used for that example) —
  left for a follow-up.

## Test plan
- Verified each corrected value by recomputing from the unchanged
  `totalSendingAmount`/`feesIncluded` using `(sendingAmount − fees) /
  exchangeRate = receivingAmount`, matching #631's basis-check method
- Cross-checked against live sandbox `/quotes` and `/exchange-rates`
  responses (2026-07-02), both consistent with this direction
- `npx mint broken-links` — clean

Fixes PROD-50 (remaining scope from #631)

### DIFF
--- a/mintlify/platform-overview/core-concepts/currencies-and-rails.mdx
+++ b/mintlify/platform-overview/core-concepts/currencies-and-rails.mdx
@@ -71,12 +71,12 @@ POST /quotes
     "decimals": 2
   },
   "totalSendingAmount": 100000,
-  "totalReceivingAmount": 1725000,
-  "exchangeRate": 17.25,
+  "totalReceivingAmount": 1716375,
+  "exchangeRate": 0.05797101,
   "feesIncluded": 500,
   "transactionId": "Transaction:019542f5-b3e7-1d02-0000-000000000025"
 }
 ```
 </Accordion>
 
-This shows: Send $1,000 → Receive $17,250 (at rate 17.25), $5 fee.
+This shows: Send $1,000 → Receive $17,163.75 (1 USD ≈ 17.25 MXN, or exchangeRate: 0.05797101), $5 fee.
--- a/mintlify/platform-overview/core-concepts/quote-system.mdx
+++ b/mintlify/platform-overview/core-concepts/quote-system.mdx
@@ -95,14 +95,14 @@ curl -X POST https://api.lightspark.com/grid/2025-10-13/quotes \
     "decimals": 8
   },
   "totalSendingAmount": 100000,
-  "totalReceivingAmount": 829167,
-  "exchangeRate": 0.00000833,
+  "totalReceivingAmount": 828835,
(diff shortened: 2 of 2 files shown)


Ethan-Minghao:
Noticed this appears related to PROD-50 per #631 — if there's already active work on this internally, happy to have this closed/redirected, just wanted to contribute since I came across the gap

shreyav (approved):
thank you!!

PR #652, opened 2026-07-06
Add platform transaction fee config + platformFees fields to the spec

jacklatourette:
## Summary

The webdev sparkcore stack shipped platform transaction fees against a hand-wired vendored Grid API client (grid-api 0.0.82). The fields are live and testable there, but they do not exist in this spec repo — so the next SDK regen from upstream would silently delete integrator-facing fields. This PR adds them to the spec source (`openapi/components/schemas/`) and regenerates the bundles (`openapi.yaml`, `mintlify/openapi.yaml`) via `npm run build:openapi`.

### 1. New `TransactionFeeConfig` schema (backs lightsparkdev/webdev#29233)

`openapi/components/schemas/config/TransactionFeeConfig.yaml` + `TransactionFeeEventType.yaml`:

- `id` (string, readOnly) — system-generated identifier
- `feeEventType` (enum: `CROSS_CURRENCY_TRANSACTION`, `TRANSFER_OUT_TRANSACTION`; only the former is accepted by `PATCH /config` today, reads return the full set)
- `sourceCurrency` (string) — ISO 4217 / crypto ticker code
- `variableFeeBps` (integer, 0–10000)
- `fixedFee` (`CurrencyAmount`) — must match `sourceCurrency`
- `isActive` (boolean, readOnly)

Required: `feeEventType`, `sourceCurrency`, `variableFeeBps`, `fixedFee`. Rows are keyed by `(feeEventType, sourceCurrency)`; deactivation is done by upserting the same key with zero fees.

### 2. `transactionFeeConfigs` on platform config (backs lightsparkdev/webdev#29233)

- `PlatformConfig.transactionFeeConfigs` — read side; returns rows regardless of `feeEventType` so operator-created rows (e.g. Transfer Out) are visible
- `PlatformConfigUpdateRequest.transactionFeeConfigs` — merge-by-key upsert; only `CROSS_CURRENCY_TRANSACTION` + `USD` accepted today, other values return 400; omit to leave configs unchanged

### 3. Platform fee surfacing on quotes and transactions (backs lightsparkdev/webdev#29402)

- `Quote.platformFeesIncluded` (int64, min 0) — the portion of `feesIncluded` collected by the platform; already included in `feesIncluded`
- `OutgoingTransaction.platformFees` (int64, min 0) — same, relative to `fees`

Descriptions match the vendored client/spec in webdev exactly, so a regen from this spec is a no-op against the hand-wired models.

## Test plan

- `npm run lint` (Redocly bundle + lint, Spectral `--fail-severity=error`) — exit 0. Only new finding is one non-failing `schema-properties-have-descriptions` warning on `TransactionFeeConfig.fixedFee`, an artifact of the `$ref` + sibling-description pattern already used throughout the spec (e.g. `embeddedWalletConfig`, `sentAmount`).
- Verified the bundled diff is exactly the new fields plus Redocly relocating `Currency`/`CurrencyAmount` earlier (now referenced from the config section).
- The webdev stack currently carries these fields in its vendored spec (grid-api 0.0.82), so the API surface is exercised end-to-end there before this merges.


jacklatourette:
Amended per webdev AT-5755: `variableFeeBps` on `TransactionFeeConfig` is now a decimal string (`pattern: ^(0|[1-9]\d*)(\.\d+)?$`, `maxLength: 255`, e.g. `"12.5"`) instead of an integer, to support fractional basis points. The ≤10000 range remains server-enforced.

Sparkcore-side change: lightsparkdev/webdev#30070.

jacklatourette:
Synced with the shipped server contract (webdev#29233): renamed to `FeeConfig`/`FeeEventType`, field `feeConfigs`, integer whole-bps `variableFeeBps`, and dropped `TRANSFER_OUT_TRANSACTION` from the enum (transfers migrate to send operations).

openapi/components/schemas/config/FeeConfig.yaml
```diff
@@ -0,0 +1,44 @@
...
+description: >-
+  A fee collected by your platform on top of Grid's fees. Keyed uniquely by
+  `(feeEventType, sourceCurrency)` within a platform. To deactivate a fee,
+  send the same key with `variableFeeBps: 0` and `fixedFee.amount: 0`.
+properties:
+  id:
+    type: string
+    description: System-generated unique identifier.
+    readOnly: true
+    example: GridPlatformTransactionFeeConfig:019542f5-b3e7-1d02-0000-000000000007
```
gcamata: let's actually leave id out of the properties (mirroring how we show platform currencies in the platform config). otherwise it could get confusing to the integrator bc the ID will change on every edit (these fee configs are append-only)

openapi/components/schemas/config/FeeConfig.yaml
```diff
@@ -0,0 +1,44 @@
...
+  feeEventType:
+    $ref: ./FeeEventType.yaml
```
gcamata: sorry, I think the eng design doc might be slightly out of sync with the codebase, but can we actually name this just "fee type" instead of "fee event type" to align with the internal name for this enum on the fee config ents?

openapi/components/schemas/config/FeeConfig.yaml
```diff
@@ -0,0 +1,44 @@
...
+      Currency code of the sending side this fee applies to. Only `USD` is
+      accepted today; other currencies return a `NOT_IMPLEMENTED` error.
+    example: USD
+  variableFeeBps:
+    type: integer
+    minimum: 0
+    maximum: 10000
+    description: >-
+      Variable fee in basis points (1 bps = 0.01%). Multiplied by the
+      transaction's source-currency amount.
```
gcamata:
```suggestion
      Variable fee in basis points (1 bps = 0.01%) to apply to a transaction's source-currency amount.
```
just a nit bc it sounds like they need to have multiplied this value at first glance

openapi/components/schemas/config/FeeConfig.yaml
```diff
@@ -0,0 +1,44 @@
...
+    example: 30
+  fixedFee:
+    $ref: ../common/CurrencyAmount.yaml
+    description: >-
+      Fixed fee charged per transaction, in the smallest unit of
+      `sourceCurrency`. `fixedFee.currency.code` must match `sourceCurrency`.
```
gcamata:
```suggestion
      Fixed fee charged per transaction. The fixed fee currency must match the fee config's source currency`.
```

openapi/components/schemas/config/FeeConfig.yaml
```diff
@@ -0,0 +1,44 @@
...
+  isActive:
+    type: boolean
+    description: >-
+      Whether this fee config is currently active. Reads return only active
+      configs.
+    readOnly: true
+    example: true
```
gcamata: nit, but do we need to return this at all if it's always true

openapi/components/schemas/config/PlatformConfigUpdateRequest.yaml
```diff
@@ -17,3 +17,13 @@ properties:
...
+  feeConfigs:
+    type: array
+    items:
+      $ref: ./FeeConfig.yaml
+    description: |
+      Merge-by-key upsert of platform fee configs, keyed by
+      `(feeEventType, sourceCurrency)`. Sending zero rates
+      (`variableFeeBps: 0` and `fixedFee.amount: 0`) for an existing key
+      deactivates it. Only `sourceCurrency: USD` is accepted today.
+      Omit this field to leave fee configs unchanged.
```
gcamata:
```suggestion
      Merge-by-key upsert of platform fee configs, keyed by (feeType, sourceCurrency)`.
      Setting variable and fixed fees to 0 for an existing fee config deactivates it.
      Omit this field to leave fee configs unchanged.
```
just making things a bit more human friendly

openapi/components/schemas/config/FeeEventType.yaml
```diff
@@ -0,0 +1,9 @@
+type: string
+enum:
+  - CROSS_CURRENCY_TRANSACTION
+description: |
+  The kind of activity this fee applies to.
+
+  - `CROSS_CURRENCY_TRANSACTION` — fee charged on a cross-currency Grid send
```
gcamata:
```suggestion
  - `CROSS_CURRENCY_TRANSACTION` — fee charged on a cross-currency Grid transaction
```

openapi/components/schemas/config/FeeConfig.yaml
```diff
@@ -0,0 +1,44 @@
+type: object
+description: >-
+  A fee collected by your platform on top of Grid's fees. Keyed uniquely by
+  `(feeEventType, sourceCurrency)` within a platform. To deactivate a fee,
+  send the same key with `variableFeeBps: 0` and `fixedFee.amount: 0`.
```
gcamata:
```suggestion
  A platform-configured fee collected by Grid and settled to the platform internal account. There can be at most one fee config for a given fee type and source currency pair. The fee will apply to all transactions of the fee type that originate in the source currency .
```

gcamata (commented):
just minor quick things

openapi/components/schemas/config/FeeConfig.yaml
jacklatourette: removed :+1:

openapi/components/schemas/config/FeeConfig.yaml
jacklatourette: removed :+1:

openapi/components/schemas/config/FeeConfig.yaml
jacklatourette: renamed to `feeType` / `FeeType` :+1:

openapi/components/schemas/config/FeeConfig.yaml
jacklatourette: applied, kept the smallest-unit note :+1:

openapi/components/schemas/config/PlatformConfigUpdateRequest.yaml
jacklatourette: applied — restored the "only USD accepted today" sentence and fixed a stray backtick :+1:

openapi/components/schemas/config/TransactionFeeConfig.yaml
```diff
@@ -0,0 +1,46 @@
...
+    description: >-
+      Whether this row is currently active. Reads return only active rows
+      unless explicitly filtered otherwise.
+    readOnly: true
+    example: true
+required:
+  - feeEventType
+  - sourceCurrency
+  - variableFeeBps
+  - fixedFee
```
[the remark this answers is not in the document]
jacklatourette: Resolved: `id` and `isActive` were removed from the schema, so no readOnly fields remain in the shared read/write schema.

openapi/components/schemas/quotes/Quote.yaml
```diff
@@ -87,6 +87,16 @@ properties:
...
+  platformFeesIncluded:
+    type: integer
+    format: int64
+    description: >-
+      The portion of `feesIncluded` collected by the platform
+      (platform-configured transaction fees), in the smallest unit of the
+      sending currency. 0 when the platform has no applicable fee configured.
+      Already included in `feesIncluded`.
+    minimum: 0
+    example: 5
```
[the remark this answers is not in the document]
jacklatourette: Added `platformFeesIncluded` to `required` :+1:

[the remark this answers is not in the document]
jacklatourette: Amendment: reverted — requiring it flagged as a breaking change for agent-action webhook consumers, so it stays optional with a note that older payloads may omit it.

gcamata (approved):

PR #683, opened 2026-07-13
docs(sca): document Strong Customer Authentication for EU customers

jklein24:
## Summary

A full, EU-scoped **Strong Customer Authentication** guide in the Mintlify docs, covering the SCA surface added in #558/#600 (both merged). What began as a single per-transaction snippet is now a dedicated guide section covering the whole SCA surface the API exposes, mirroring the structure of the reference docs it wraps ([Striga's SCA docs](https://docs.striga.com/reference/strong-customer-authentication-sca.md)).

## What's added

A new **"Strong Customer Authentication"** group under the *Get started* tab (`docs.json`), with six pages under `platform-overview/sca/`:

- **`overview`** — region scoping (EU / EUR / USDC), a "what SCA covers" table, factors + dynamic linking, a Mermaid flow diagram of the authorize loop, and **lifetimes/limits** + **error** tables.
- **`per-transaction-authorization`** — reuses the existing `snippets/sca/strong-customer-authentication.mdx` as the single source (the three money-movement flow pages still import the same snippet, so they're unchanged): `PENDING_AUTHORIZATION` + `scaChallenge`, the status-driven authorize loop, inline vs. after-the-fact proof, resend, realtime-funding quotes, sandbox `123456`.
- **`factor-enrollment`** — TOTP + passkey enrollment (start/confirm), factor list/delete.
- **`login-and-sessions`** — SCA login + the session it grants, plus account-security (`record-event`) signals and the lockout ladder.
- **`trusted-beneficiaries`** — trust and untrust as start → confirm (keyed by `externalAccountId`).
- **`two-factor-reset`** — liveness-gated start → poll → complete.

## Refreshed for API drift (2026-08-28)

Rebased onto current `main` — the branch previously carried #600's spec commits, which are merged, so the PR is now **docs-only** — and updated for everything that landed on the SCA surface since the guide was written:

- **Login complete** requires `endUserIpAddress` and returns `sessionExpiresAt` (#780); the session guidance now tells integrators to prompt a re-login ahead of the expiry instead of discovering it on a failed call.
- **Quote authorize** documents `409 SCA_SESSION_REQUIRED` and `423 ACCOUNT_LOCKED` (#761); both join the error tables, and the snippet notes that authorizing requires an active login session.
- **Abandoned challenges** now fail the transaction with `failureReason: SCA_NOT_COMPLETED` and no funds moved (#762).
- **Trusted external accounts** refuse `DELETE` with `409 BENEFICIARY_TRUSTED` — untrust (SCA-gated) before deleting (#770).
- The challenge lives on the **quote**, not the transaction; webhook consumers route via the transaction's `quoteId` (#701). Fixed the "transaction (or quote) … `scaChallenge`" phrasing in the snippet, the flow-page blurbs, and the Mermaid authorize label.
- **`POST /transfer-out` is deprecated** in favor of `POST /quotes` with `immediatelyExecute: true` (#856); the transfer-out tab now says so.
- Corrected the suspension guidance to match the spec: a suspended customer clears the lockout by completing a password reset (`RESET_PASSWORD_COMPLETED`).

## Review addressed

Incorporated the review on the guide:

- Bounded the 2FA-reset poll loop (stop at `expiresAt` / client timeout; restart if the window closes before liveness passes).
- Clarified the failed-login counter is cumulative and cleared only by `RESET_PASSWORD_COMPLETED`, not by a successful login.
- Untrust is documented as start → confirm, backed by the `startBeneficiaryUntrust` endpoint from #600.
- Fixed MD031/MD040 markdownlint violations on the new fenced code blocks.

Prose pass: em-dashes and AI-tell phrasing removed across the pages and the shared snippet.

## Validation

- `make build` — rebundled `openapi.yaml` / `mintlify/openapi.yaml` byte-match `main`; the PR carries no spec changes.
- `make lint` (redocly + spectral) passes.
- `markdownlint` (repo-pinned `markdownlint-cli`) passes clean on all SCA pages and the touched flow pages.
- `mint broken-links` adds **zero** new broken links versus `main`.
- Two adversarial-review rounds: no P0/P1 findings; P2s fixed (authorize-call label in the Mermaid diagram, suspension-vs-password-reset wording, `PENDING_AUTHORIZATION` guard in the conversion example, `type`/`factor` discriminator wording).

### DIFF
--- a/mintlify/payouts-and-b2b/payment-flow/send-payment.mdx
+++ b/mintlify/payouts-and-b2b/payment-flow/send-payment.mdx
@@ -264,6 +264,15 @@ end-to-end times measured from quote execution:
   of these.
 </Info>
 
+## Strong Customer Authentication (EU customers)
+
+Customers in SCA-regulated regions (in practice the EU: EUR / USDC) must confirm
+payments with Strong Customer Authentication. When it applies, the payment comes
+back `PENDING_AUTHORIZATION` with an `scaChallenge` that you authorize before the
+transfer is released; for every other customer nothing changes. See
+[Per-transaction authorization](/platform-overview/sca/per-transaction-authorization)
+for the full walkthrough.
+
 ## Checking Payment Status
 
 Configure a webhook endpoint to receive real-time notifications when payment status changes:
--- a/mintlify/payouts-and-b2b/payment-flow/send-payment.mdx
+++ b/mintlify/payouts-and-b2b/payment-flow/send-payment.mdx
@@ -267,9 +267,9 @@ end-to-end times measured from quote execution:
 ## Strong Customer Authentication (EU customers)
 
 Customers in SCA-regulated regions (in practice the EU: EUR / USDC) must confirm
-payments with Strong Customer Authentication. When it applies, the payment comes
-back `PENDING_AUTHORIZATION` with an `scaChallenge` that you authorize before the
-transfer is released; for every other customer nothing changes. See
+payments with Strong Customer Authentication. When it applies, the quote comes
+back `PENDING_AUTHORIZATION` carrying an `scaChallenge` that you authorize
+before the transfer is released; for every other customer nothing changes. See
 [Per-transaction authorization](/platform-overview/sca/per-transaction-authorization)
 for the full walkthrough.
 
--- a/mintlify/platform-overview/sca/login-and-sessions.mdx
+++ b/mintlify/platform-overview/sca/login-and-sessions.mdx
@@ -1,0 +1,123 @@
+---
+icon: "/images/icons/shield.svg"
+title: "Login & sessions"
+description: "The end-user SCA login and the session it grants"
+"og:image": "/images/og/og-get-started.png"
+---
+
+<Note>
+Applies only to customers in an SCA-required region (EU). Every endpoint here
+returns **`409`** for other customers.
+</Note>
+
+Per-transaction authorization gates individual debits. The **SCA login** is
+separate: it authenticates the end user to open a longer-lived session that
+covers reads and account access beyond the per-transaction window. Grid provides
+the login plumbing; your application decides when to drive it (for example, when
+a customer opens their account and the previous session has lapsed).
+
+<Note>
+**A customer's EUR / USDC accounts aren't provisioned until their first SCA
+login after KYC approval.** Provisioning is deferred from KYC-approval time to
+the first login that opens a valid SCA session, so a freshly KYC-approved
+customer's EUR / USDC accounts won't appear in `GET /customers/internal-accounts`
+until then. Expect those accounts to be unavailable, and drive the SCA login
+once KYC is approved, before relying on them.
+</Note>
+
+All paths below are relative to `https://api.lightspark.com/grid/2025-10-13`.
+
+## Logging in
+
+<Steps>
+<Step title="Start the login">
+
+```bash
+POST /sca/login/start?customerId={customerId}
+
+{ "factor": "SMS_OTP" }
+```
+
+The response carries only what the chosen factor needs:
+
+- **`SMS_OTP`**: a code is dispatched; you get back `challengeId` and `expiresAt`.
+- **`TOTP`**: nothing extra; the customer reads the code from their app.
+- **`PASSKEY`**: WebAuthn `passkeyOptions` (with `allowedOrigins` and `relyingPartyId`) to pass to the device.
+
+The factor must already be enrolled (or, for `SMS_OTP`, the phone verified). See
+[factor enrollment](/platform-overview/sca/factor-enrollment).
+</Step>
+<Step title="Complete the login">
+Submit the proof for the factor you started with: `code` for `SMS_OTP` / `TOTP`
+(echoing `challengeId` for `SMS_OTP`), or `passkeyAssertion` + `origin` for
+`PASSKEY`.
+
+```bash
+POST /sca/login/complete?customerId={customerId}
+
+{ "factor": "SMS_OTP", "challengeId": "…", "code": "123456" }
+```
+
+```json
+{ "status": "SUCCESS" }
+```
+
+A `status` of `SUCCESS` means the session is open for 180 days and revokes any
+previous SCA session for that customer. Any other value means the
+login did not complete; the field is passed through verbatim, so treat only
+`SUCCESS` as success. An invalid or expired proof returns `400`. **In sandbox,
+the code is always `123456`.**
+</Step>
+</Steps>
+
+## Session scope and fresh authentication
+
+An active SCA login session covers EUR / USDC account reads for 180 days. A
+request for transaction history older than 90 days requires fresh SCA, even when
+the broader session has not expired. When Grid indicates that a session is
+missing, expired, or too old for the requested history, restart the login flow
+before retrying the read.
+
+## Account-security signals
+
+Grid runs an adaptive-authentication risk engine that maintains each customer's
+login-security state. Because your application owns the customer's login, you
+report the security-relevant events it sees so the engine can act on them:
+
+```bash
+POST /sca/record-event?customerId={customerId}
+
+{ "eventType": "FAILED_LOGIN_ATTEMPT" }
+```
+
+Returns the customer's resulting login-security state so you can surface a
+lockout — `{ eventType, suspended, lockedUntil, failedAttempts }`. When the
+customer is locked out, this (and `POST /sca/login/complete`) returns `423` with
+`details.lockedUntil` (when they may retry) and `details.failedAttempts`.
+`eventType` must be one of:
+
+| `eventType` | Effect |
+|---|---|
+| `FAILED_LOGIN_ATTEMPT` | Increments the failed-login counter and escalates a lockout: **5 → 15 min, 6 → 30 min, 7 → 1 hour, 8 → 24 hours, 9 or more → suspension.** |
+| `RESET_PASSWORD_COMPLETED` | Revokes every active SCA session for the customer and clears the failed-login counter. |
+
+Report `FAILED_LOGIN_ATTEMPT` on each failed sign-in and `RESET_PASSWORD_COMPLETED`
+once a password recovery finishes. Any other value returns `400`.
+
+<Note>
+The failed-login counter is cumulative and is **not** reset by a successful
+login; only `RESET_PASSWORD_COMPLETED` clears it. Record that event after a
+password recovery to zero the counter and clear a time-bounded lockout, rather
+than relying on the customer simply logging in again. A suspended customer
+(9 or more failed attempts) requires support intervention; password recovery
+does not unsuspend the account.
+</Note>
+
+## Your responsibilities
+
+Grid provides the SCA endpoints and risk decisions; your application owns the
+end-user login and session experience. Report every failed sign-in and completed
+password recovery through `record-event`, enforce any returned lockout before
+offering another login attempt, and do not store or reuse a customer's TOTP
+secret or passkey material. Treat TOTP secrets and WebAuthn ceremony data as
+end-user credentials, not platform credentials.
--- a/mintlify/platform-overview/sca/login-and-sessions.mdx
+++ b/mintlify/platform-overview/sca/login-and-sessions.mdx
@@ -50,33 +50,39 @@ The factor must already be enrolled (or, for `SMS_OTP`, the phone verified). See
 <Step title="Complete the login">
 Submit the proof for the factor you started with: `code` for `SMS_OTP` / `TOTP`
 (echoing `challengeId` for `SMS_OTP`), or `passkeyAssertion` + `origin` for
-`PASSKEY`.
+`PASSKEY`. Every completion also requires **`endUserIpAddress`** — the IP of the
+end user's device the login is performed from, recorded against the login event
+and fed into risk assessment. Supply the customer's address, not your server's.
 
 ```bash
 POST /sca/login/complete?customerId={customerId}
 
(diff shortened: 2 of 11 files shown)


jklein24:
* **#683** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/683?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/683?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#600** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/600?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#558** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/558?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

jklein24:
@greptile review

mintlify/snippets/sca/strong-customer-authentication.mdx
```diff
@@ -0,0 +1,108 @@
...
+Under PSD2, EU e-money and e-money-token (EUR / USDC) money movement must be
+confirmed by the end user with Strong Customer Authentication (SCA). Grid wraps
+the provider's SCA so you satisfy it through the same resources you already use —
+there is no separate SCA product to integrate.
+
+### When you'll encounter it
+
+For an SCA-required customer, a money-movement call that would otherwise complete
+instead returns the transaction (or quote) in status **`PENDING_AUTHORIZATION`**
+with an **`scaChallenge`** object, and the transfer is **not** released until the
```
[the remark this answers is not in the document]
jklein24: **Applied.** Added a note that a realtime-funding quote's `202` withholds `paymentInstructions` until the challenge is authorized — authorize first, then read them from the returned quote, not off the initial pending response.

_Review response generated using `/greploop` Claude command._

mintlify/snippets/sca/strong-customer-authentication.mdx
```diff
@@ -0,0 +1,108 @@
...
+</Step>
+</Steps>
+
+If an SMS code lapses before it's used, re-send it (the existing challenge is
+reused, and its `expiresAt` is not extended):
+
+```bash
+curl -X POST https://api.lightspark.com/grid/2025-10-13/transactions/{transactionId}/authorize/resend \
+  -u "$GRID_API_TOKEN"
+```
```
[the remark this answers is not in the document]
jklein24: **Applied.** The resend section now lists both endpoints — `POST /transactions/{transactionId}/authorize/resend` and `POST /quotes/{quoteId}/authorize/resend` — keyed on which resource carries the challenge.

_Review response generated using `/greploop` Claude command._

jklein24:
@greptile review

jklein24:
@greptile review

mintlify/platform-overview/sca/trusted-beneficiaries.mdx
```diff
@@ -0,0 +1,73 @@
...
+```
+
+An invalid or expired proof returns `400`. **In sandbox, the code is always
+`123456`.**
+</Step>
+</Steps>
+
+## Untrusting a beneficiary
+
+Revoking trust is a single confirm call with the SCA proof, the same request
```
jklein24: Untrust is documented here as a single confirm call that expects {challengeId, code}, but I don't see anything on the untrust surface that actually issues that challenge — we only have trust, trust/confirm, and untrust/confirm; there's no untrust start endpoint, and trust start is documented as kicking off trusting, not untrusting. For an SMS_OTP-only customer (our default, no enrollment needed) there's no way to get that code, and a PASSKEY customer has no server-issued assertion options — only TOTP can self-serve here. Can we either document that trust start also issues the untrust challenge, or add an untrust start endpoint and update this page to reflect it?

mintlify/platform-overview/sca/two-factor-reset.mdx
```diff
@@ -0,0 +1,63 @@
...
+GET /customers/{customerId}/sca/factors/reset/{resetId}
+```
+
+```json
+{ "status": "PENDING" }
+```
+
+Keep polling while `status` is non-terminal (e.g. `PENDING`). Once it reaches
+`LIVENESS_PASSED`, proceed. The field is passed through verbatim, so treat any
+value other than `LIVENESS_PASSED` as "not ready yet" rather than a failure.
```
jklein24: The polling guidance contradicts itself: we say to keep polling while status is non-terminal, implying there are terminal values, but then say anything other than LIVENESS_PASSED should be treated as 'not ready yet' rather than a failure. There's no documented terminal-failure status (REJECTED/FAILED/EXPIRED) or stop condition, so a customer who terminally fails liveness leaves the integrator polling until expiresAt — and since expiresAt is optional in the start schema, that loop might not even be bounded. Let's enumerate the terminal statuses and specify when to stop polling.

mintlify/platform-overview/sca/login-and-sessions.mdx
```diff
@@ -0,0 +1,82 @@
...
+POST /customers/{customerId}/sca/record-event
+
+{ "eventType": "FAILED_LOGIN_ATTEMPT" }
+```
+
+Returns `204`. `eventType` must be one of:
+
+| `eventType` | Effect |
+|---|---|
+| `FAILED_LOGIN_ATTEMPT` | Increments the failed-login counter and escalates a lockout: **5 → 15 min, 6 → 30 min, 7 → 1 hour, 8 → 24 hours, 9 or more → suspension.** |
```
jklein24, line 109: Have you considered whether the failed-login counter resets on anything besides RESET_PASSWORD_COMPLETED? As written, the table reads like failures accumulate over the customer's lifetime regardless of successful logins in between, so 9 failures spread across years of otherwise-successful logins would still trigger suspension. Does a successful SCA login reset the counter, or does the platform need to drive a password reset to unstick customers approaching the threshold?

mintlify/platform-overview/sca/login-and-sessions.mdx
```diff
@@ -0,0 +1,82 @@
...
+
+## Logging in
+
+<Steps>
+<Step title="Start the login">
+```bash
+POST /customers/{customerId}/sca/login/start
+
+{ "factor": "SMS_OTP" }
+```
```
jklein24, line 39: Nit (optional): the new fenced code block here (and the similar one at two-factor-reset.mdx:26) is missing a blank line around the fence and a language tag, which trips MD031/MD040 in our markdownlint config. Worth fixing while it's fresh, since CI won't currently catch it.

jklein24 (commented):
Request changes

mintlify/platform-overview/sca/trusted-beneficiaries.mdx
jklein24: Good catch, and you're right that this is a surface gap rather than just a doc slip: there's no `startBeneficiaryUntrust` operation, only `startBeneficiaryTrust` plus the two confirms. I've reworked the untrust section to say so plainly: untrust has no start step, so it's confirmed with a proof the customer can produce on their own (a TOTP code) with `challengeId` omitted, and I've noted that SMS/passkey untrust would each need a start endpoint that isn't in the surface yet. Adding that endpoint is a spec + backend change beyond this docs PR, so I'm tracking it as follow-up rather than papering over it here.

mintlify/platform-overview/sca/two-factor-reset.mdx
jklein24: Agreed. The status is a passthrough string with only `PENDING` and `LIVENESS_PASSED` defined, so rather than invent terminal-failure values I bounded the loop explicitly: stop at `expiresAt` (or a client-side timeout when the start response omits it), and if the window closes before liveness passes, start a new reset instead of polling on. Updated the poll step.

mintlify/platform-overview/sca/login-and-sessions.mdx
jklein24: It's lifetime-cumulative: only `RESET_PASSWORD_COMPLETED` clears the counter, a successful login does not. I've added a note making that explicit and recommending platforms record `RESET_PASSWORD_COMPLETED` after a password recovery to zero the counter and lift a lockout.

mintlify/platform-overview/sca/login-and-sessions.mdx
jklein24: Fixed. Added the missing blank lines and language tags; `markdownlint-cli2` now passes clean across all the SCA pages.

mintlify/platform-overview/sca/trusted-beneficiaries.mdx
jklein24: Update: rather than leave this as a gap, I closed it end to end. Added `startBeneficiaryUntrust` (POST .../untrust) to the spec (#600), mirroring trust-start so it issues the SCA challenge that untrust/confirm satisfies, plus the sparkcore backend (StrigaClient initiate-untrust → wallets/beneficiary/{iban,address}/untrust, the grid-switch start method resolving the persisted whitelist handle, and the handler). This page now documents untrust as start → confirm, matching trust. So SMS_OTP/passkey untrust works now, not just TOTP.

mintlify/platform-overview/sca/per-transaction-authorization.mdx
```diff
@@ -0,0 +1,15 @@
...
+description: "Authorize an SCA-gated money movement"
+"og:image": "/images/og/og-get-started.png"
+---
+
+This is the flow you hit most often: an SCA-required customer initiates a money
+movement, it comes back `PENDING_AUTHORIZATION` with an `scaChallenge`, and you
+authorize it before the transfer is released. For where this sits in the wider
+SCA surface, see the [overview](/platform-overview/sca/overview).
+
+import StrongCustomerAuthentication from '/snippets/sca/strong-customer-authentication.mdx';
```
jklein24, line 14: I don't love just reusing the generic snippet here unless we improve that snippet to be in more of a walkthrough format. Maybe we can do that? We should really clearly show the sequence of calls for different types of flows - a realtime quote flow, an execute quote flow, and transfer-out. This would let you see the whole thing end-to-end.

mintlify/payouts-and-b2b/payment-flow/send-payment.mdx
```diff
@@ -362,6 +363,10 @@ If a transaction fails, Grid initiates a refund automatically. You'll receive `O
 | `FAILED`     | Transfer failed — refund initiated automatically (track via `refund` object) |
 | `EXPIRED`    | Quote expired without execution            |
 
+## Strong Customer Authentication (EU customers)
+
+<StrongCustomerAuthentication />
```
jklein24: I'm slightly worried that embedding this snippet on these "sending payments" pages will be super verbose in this context. Maybe instead, we can include a short blurb and a link to the SCA docs section?

mintlify/platform-overview/sca/per-transaction-authorization.mdx
jklein24: **Applied.** Reworked this page so it's no longer just the generic snippet: added a **Walkthrough by flow** section with a tab per flow — *execute a quote*, *realtime-funding quote*, and *transfer-out* — each laying out the exact call sequence end to end (which call first returns the `scaChallenge`, and which resource you authorize). The shared snippet stays above it as the concept/factor reference.

mintlify/payouts-and-b2b/payment-flow/send-payment.mdx
jklein24: **Applied.** Dropped the full snippet from the flow pages — this one, plus `global-p2p/sending-payments` and `ramps/fiat-crypto-conversion` — and replaced it with a two-sentence blurb + a link to the [per-transaction authorization](/platform-overview/sca/per-transaction-authorization) walkthrough. Keeps these pages focused while the SCA detail lives in one place. (Applied to all three flow pages for consistency; shout if you'd rather leave the conversion page verbose.)

jklein24:
Implementation follow-ups from the Striga SCA parity review (docs-only corrections are in b66552e):

1. **Operation-aware factor support.** Striga permits TOTP for non-dynamically-linked crypto withdrawals and trusted-beneficiary EUR/USDC sends, while dynamically linked debits require SMS/passkey. Grid needs to derive `availableFactors` from operation, destination, and trust state, expose a selectable factor where needed, and add contract tests for BTC/Lightning, trusted USDC, and untrusted regulated debits. Trust/untrust start must also accept a factor, or be documented and enforced as SMS-only; otherwise callers cannot initiate a passkey ceremony.

2. **2FA reset lifecycle.** Grid returning `204` from completion plus independent re-enrollment does not model Striga factor-specific completion results, such as new TOTP material or passkey creation options, or the transition to `COMPLETED` only after enrollment confirmation. Either expose that continuation or prove the alternate flow with contract tests and a coherent status model; include pending-reset reuse, enrolled-factor mismatch, suspended-user recovery, and session-revocation behavior.

3. **Strict proof contracts.** Make authorization, login completion, and beneficiary-confirm schemas discriminated variants: exactly one proof, passkey requires assertion plus origin, and SMS requires challenge ID. Add negative validation tests so `{}` and invalid factor/proof combinations fail at schema validation.

4. **Session enforcement.** Confirm or implement the 180-day EUR/USDC read session, fresh SCA for history older than 90 days, and prior-session revocation on login. Surface deterministic Grid errors on the affected read endpoints.

Reference: https://docs.striga.com/reference/strong-customer-authentication-sca and https://docs.striga.com/reference/sca-factors-implementation.

shreyav (approved):

PR #578, opened 2026-06-12
Add stablecoin issuance API surface

kphurley7:
> [!IMPORTANT]
> **Do not merge until the webdev implementation stack (lightsparkdev/webdev#28785 → #28734) has merged and the feature is ready to light up.** Merging here publishes the endpoints: the nightly docs-sync workflow picks up `openapi/` changes on main and the Stainless-documented spec feeds the public docs/SDKs. This PR stays draft until launch readiness.

## Summary

Companion spec PR for the Brale stablecoin issuance stack in webdev (implementation: the flow-sliced stack starting at lightsparkdev/webdev#28785). Adds to the split OpenAPI source (`openapi/`):

- **12 paths** under `openapi/paths/stablecoins/`: provider account links, stablecoin registration/list/get, external-account link sessions + linking (Plaid / direct bank entry / Grid external account), mint+burn quote create/get/execute, operations list/get.
- **48 schemas** under `openapi/components/schemas/stablecoins/`: typed discriminated requests (quote operation types, mint funding sources, link methods), typed `destination` oneOf on quote/operation, payout-rail subset enum, list envelopes.
- **Stablecoin error codes** added to `Error400`/`Error404`/`Error409` enums (the generated clients hard-reject unknown codes, so these must ship before the backend returns them).
- `Stablecoins` tag.

Bundle regenerated with `make build`; Redocly validation clean; spectral passes at `--fail-severity=error`.

## Verification

The rebuilt bundle's stablecoin paths/schemas and error enums are deep-equal to the spec vendored in the webdev stack tip (verified programmatically). The only non-stablecoin delta vs that copy is `ExchangeRateFees.total`, which exists here and not yet in webdev's vendored copy (pre-existing drift; webdev picks it up on its next `update_schema.sh` sync).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #687, opened 2026-07-14
Add stablecoin direct mint/burn issuance API surface

kphurley7:
## What

Adds the public API surface for **direct stablecoin issuer operations** that shipped in the Grid server (webdev #29077 mint, #29078 burn). Upstream `main` currently has only the stablecoin *provider-account* surface (from #594); the mint/burn/registration surface lives only in webdev's vendored `grid-api/openapi.yaml` and is absent here — so it gets clobbered on the next `update_schema.sh` regen and external SDK consumers never receive it.

## Surface added

- **Paths:** `/stablecoins`, `/stablecoins/{stablecoinId}`, `/stablecoins/{stablecoinId}/mints`, `/burns`, `/operations`, `/stablecoin-operations/{stablecoinOperationId}`.
- **Schemas** (`components/schemas/stablecoins/`): `Stablecoin`, `StablecoinListResponse`, `StablecoinRegisterRequest`, `StablecoinMintRequest`, `StablecoinBurnRequest`, mint/burn source + destination, `StablecoinOperation`, `StablecoinOperation{Type,Status,Destination,ListResponse}`, `StablecoinFundingInstructions`, `StablecoinEstimatedDelivery`, `StablecoinNetwork`, `Stablecoin{Grid,}OperationsStatus`, `StablecoinIssuanceStatus`.
- Adds `Error503` (SERVICE_UNAVAILABLE) used by the issuance responses.

## Note on `StablecoinOperation.destination`

It is a single merged `StablecoinOperationDestination` object (superset of the mint/burn request destinations), **not** a `oneOf`. A non-discriminated `oneOf[Mint,Burn]` is ambiguous — a burn destination validates as both variants — so the generated SDK raises "Multiple matches found" on every burn operation response. The merged object deserializes unambiguously.

## Supersedes #578

#578 modeled issuance as **quotes** (`/quotes` + `execute`); the shipped server implements **direct** mint/burn operations. This PR reflects what shipped. #578 is being closed.

## Validation

Built with `npm run build:openapi` (redocly bundle) and `redocly lint` passes. Draft pending review; a follow-up regenerates webdev's vendored client from this once merged.


openapi/components/schemas/stablecoins/StablecoinFundingInstructions.yaml
```diff
@@ -0,0 +1,8 @@
+type: object
+description: Provider funding instructions safe to show publicly. Present for wire-funded mint operations and external-source burn operations when available.
+additionalProperties: true
+example:
+  rail: SPARK
+  network: SPARK
+  address: spark1...
+  valueType: ACME
```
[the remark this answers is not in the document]
shreyav: I agree, is it possible to make this typed? or is it intentionally flexible

openapi/components/schemas/stablecoins/StablecoinEstimatedDelivery.yaml
```diff
@@ -0,0 +1,8 @@
+type: object
+description: Static rail timing estimate for fiat redemption delivery. This is guidance, not a guaranteed arrival time.
+additionalProperties: true
+example:
+  rail: ACH_CREDIT
+  amount: '1.00'
```
shreyav: should this also be minor units

openapi/components/schemas/stablecoins/StablecoinMintDestination.yaml
```diff
@@ -0,0 +1,18 @@
+type: object
+required:
+  - type
+properties:
+  type:
+    type: string
+    enum:
+      - EXTERNAL_ACCOUNT
```
shreyav: do we need this distinction? on the other apis we treat internal and external accounts as an "account" type so it might be nice for consistency to keep that pattern

shreyav: for example quote destination https://docs.lightspark.com/api-reference/cross-currency-transfers/create-a-transfer-quote#body-destination

openapi/components/schemas/stablecoins/StablecoinMintFundingSource.yaml
```diff
@@ -0,0 +1,29 @@
...
+    description: Grid internal funding account id for future Grid-funded mint flows.
+    example: InternalAccount:019542f5-b3e7-1d02-0000-000000000102
+  sourceTokenIdentifier:
+    type: string
+    description: Provider token/value type for future provider internal balance funding.
+    example: USDC
+  cryptoNetwork:
+    type: string
+    description: Source crypto network for future provider internal balance funding.
+    example: SOLANA
```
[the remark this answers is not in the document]
shreyav: would it be possible to use a similar pattern to quotes here as well? the one of pattern would give us that schema level enforcement

openapi/components/schemas/stablecoins/StablecoinOperationDestination.yaml
```diff
@@ -0,0 +1,33 @@
...
+required:
+  - type
+properties:
+  type:
+    type: string
+    enum:
+      - EXTERNAL_ACCOUNT
+      - GRID_INTERNAL_ACCOUNT
+    description: Destination variant.
+  externalAccountId:
```
shreyav: same comment as above on using account id, or a one of if we anticipate having other destination types

openapi/components/schemas/stablecoins/StablecoinFundingInstructions.yaml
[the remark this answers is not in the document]
kphurley7: Made this typed in 3afced11. `StablecoinFundingInstructions` now declares its stable core fields (`rail`, `network`, `address`, `valueType`) with types and examples, and `StablecoinEstimatedDelivery` is fully typed as well. The only reason `additionalProperties: true` stays is to pass through the rail/provider-specific remainder without a spec change per rail (e.g. wire instructions carry bank beneficiary/account/routing details, while Spark/on-chain instructions carry a deposit address). So the common shape is typed and the variable tail stays open by design.

openapi/components/schemas/stablecoins/StablecoinEstimatedDelivery.yaml
kphurley7: Yes, addressed in 3afced11. `StablecoinEstimatedDelivery.amount` is now a string in the smallest unit of `currency` (e.g. cents for USD, `pattern: ^[0-9]+$`), matching the convention used elsewhere in this spec (`CurrencyAmount`, `StablecoinMintRequest.amount`, etc.).

openapi/components/schemas/stablecoins/StablecoinMintDestination.yaml
kphurley7: Agreed, unified in 227eedb9. `StablecoinMintDestination` is now a single object with one required `accountId`, dropping the `oneOf` + `type: EXTERNAL_ACCOUNT | GRID_INTERNAL_ACCOUNT` discriminator (and its `StablecoinExternalAccountMintDestination` / `StablecoinGridInternalMintDestination` / `StablecoinMintDestinationBase` variants). This matches the existing `AccountDestination` / `AccountQuoteSource` pattern you linked, where the id prefix (`ExternalAccount:` vs `InternalAccount:`) disambiguates internal vs external. It also removes the conditionally-required-fields concern, since a single always-required field has no conditional case.

openapi/components/schemas/stablecoins/StablecoinOperationDestination.yaml
kphurley7: Same unification applied on the response side in 227eedb9. `StablecoinOperationDestination` now exposes a single `accountId` (merging the former `externalAccountId` / `accountId` fields and dropping the redundant external/internal `type`, since the id prefix carries that distinction). I kept it as a flat, non-`oneOf` merged shape so the field stays unambiguously deserializable regardless of operation type: `rail` is present for burn (fiat payout) destinations and absent for mint. I did not add a `oneOf` for other destination kinds since there is no non-account destination type in this domain yet; we can introduce one if that changes.

shreyav (approved):

shreyav (approved):

PR #700, opened 2026-07-18
plan-kyc-api-design

akanter:
## Summary

Design proposal for the API-based KYC flow — an additive extension of the existing KYB verification surface. Full doc: `docs/plans/2026-07-18-kyc-api-design.md` in this PR's diff.

The direct-API KYC flow already technically works today (`POST /customers` INDIVIDUAL → `POST /documents` → `POST /verifications` returns structured `errors[]`). This document proposes the additions to make that flow production-ready for **interactive in-app onboarding**. Bulk migration is documented in **Appendix A** for continuity but is explicit future work — not on this workstream.

## Reading guide

The doc is structured to be read top-to-bottom for the singleton flow (Phase 1). Bulk migration is at the end (Appendix A), skippable.

1. **Context** — where we are today.
2. **Design goals** — the axes we're optimizing.
3. **Open questions for Phase 1** — decisions needing input, each with a proposed default.
4. **What exists today** — API surface + backend behavior + existing sandbox pattern.
5. **Proposed changes** — §1 `taxIdentification`, §1a EDD endpoint, §2 400-level validation, §3 Document upload strategy, §4 Sandbox mode.
6. **Walkthrough** — interactive singleton flow example.
7. **Implementation phasing** — Phase 1 (API + sandbox) → Phase 2 (production wire-up).
8. **Appendix A** — Bulk migration (deferred).

## Open questions calling for input

Three decisions the doc surfaces near the top (each with a proposed default so silence = proceed):

1. **`/customers/edd` vs `/customers/endorsements` (Bridge-style)?** Jeremy raised this via Aaron. Bridge uses endorsement-per-rail (`base`/`sepa`/`spei`/`pix`/`faster_payments`/`cop`) with independent lifecycles; Stripe uses a single per-customer requirements bag. Current doc uses `/customers/edd` (single bag). **Default: keep for Phase 1, pivot before the second compliance regime lands.**
2. **`PurposeOfAccount` — need an individual variant?** Existing enum values are business-flow oriented. **Default: add `IndividualPurposeOfAccount`, matching the pattern applied to `IndividualSourceOfFundsCategory`.**
3. **EDD backend storage — JSON blob vs dedicated ent?** **Default: dedicated `EntUmaaasCustomerEdd` ent.**

## Approach

Keep KYC on the same `Verification` lifecycle as KYB. All extensions are additive — no breaking changes to `Customer`, `Verification`, or `VerificationError`. Universal wire format for addresses; per-jurisdiction / per-currency requirements are documented in Mintlify and surfaced reactively via `POST /verifications` errors (no runtime discovery endpoint). Sandbox is first-class — platforms drive the full KYC lifecycle (create → verify → APPROVED / REJECTED / PENDING / EDD-required) via deterministic `fullName`-suffix magic values that mirror the existing KYB and BO KYC patterns.

Implementation strategy: ship the API surface + sandbox behavior first (customers can integrate immediately); production wire-up follows behind the same API.

## Proposed changes (Phase 1)

- **§1 `taxIdentification`** on `IndividualCustomerFields` — single optional object of `{idType, identifier, countryOfIssuance}`, using existing `IdentificationType` enum.
- **§1a EDD endpoint** — new `POST/GET/PATCH /customers/edd` (customerId in body on POST, query param on GET/PATCH), matching the flat-linked-resource pattern from `openapi/README.md`. Schema: 8 primary EDD fields + 3 conditional `*OtherDescription` freeform fields.
- **§2 Machine-readable 400 validation** — new `FieldConstraint.yaml` under `errors/`, documents the `Error400.details.constraint` shape (regex / format / enum / min-max length). Applies to `POST /customers`, `PATCH /customers/{id}`, `POST /customers/edd`, `PATCH /customers/edd`.
- **§3 Document upload strategy** — Patterns A (no docs) + B (sequential create-then-upload). Both are single-customer. Bulk-oriented document patterns (C, D) live in Appendix A.
- **§4 Sandbox** — individual-KYC magic-value triggers on `fullName` suffix: `001` PENDING, `002` REJECTED, `003` EDD-required, else APPROVED. Documented alongside the existing business KYB and BO KYC sandbox in `mintlify/snippets/sandbox-verification.mdx`.

## Not in this workstream

Bulk migration + Grid-side document caching (Appendix A). OFAC / LexisNexis handling is a separate work item.

Reply with a comment (e.g. LGTM) to approve the design direction — emoji reactions don't notify me here.

### DIFF
--- a/docs/plans/2026-07-18-kyc-api-design.md
+++ b/docs/plans/2026-07-18-kyc-api-design.md
@@ -1,0 +1,345 @@
+# API-based KYC flow — design proposal
+
+## Context
+
+Grid platforms can already onboard individual customers today via the direct-API flow: `POST /customers` (INDIVIDUAL) → `POST /documents` → `POST /verifications`. The verification response already carries a structured `errors[]` array with `MISSING_FIELD` / `INVALID_FIELD` / `MISSING_*_DOCUMENT` entries.
+
+This document proposes the additions needed to make that flow production-ready for **interactive in-app onboarding** — a platform collects data over multiple rounds as the end-user fills in forms. Bulk migration of pre-verified customers is covered in **[Appendix A](#appendix-a--bulk-migration-deferred-future-work)** and is not on the current workstream.
+
+The design is additive to Victor's existing KYB verification surface. KYC and KYB share the same `Verification` lifecycle; the changes here fill gaps in the individual-customer schema and add pieces the existing surface doesn't cover.
+
+**Reference:** [Lead Bank KYC requirements spreadsheet](https://docs.google.com/spreadsheets/d/1pxyK7j8k865u_ZxUp1Ymjdigyo76JK7tAzsetUnP8oQ/edit?gid=1402608033#gid=1402608033) — canonical spec for individual retail KYC field requirements addressed here (Lightspark internal).
+
+## Design goals
+
+- One API supports both interactive and (future) bulk usage.
+- Missing or invalid data is surfaced at two layers: **format-level** issues (regex, enum, length) surface as `400 INVALID_INPUT` at submit time with structured `details.constraint`, so the platform's onboarding form can pre-validate. **Verification-level** issues (missing required fields for this jurisdiction, Sumsub-side rejections, document quality) surface as `errors[]` on `POST /verifications`, categorized by the existing `VerificationErrorType` enum.
+- Country / nationality / currency variations are handled via documented per-jurisdiction requirements plus reactive validation errors, not schema forks or a runtime discovery endpoint.
+- Additive-only — no breaking changes to `Customer`, `Verification`, or `VerificationError`.
+- **Sandbox is first-class.** Platforms can drive the full KYC lifecycle in sandbox (create → verify → APPROVED / REJECTED / PENDING / EDD-required) without hitting real vendors, using deterministic magic values that mirror the existing KYB and beneficial-owner-KYC sandbox patterns.
+
+## Open questions for Phase 1
+
+Design decisions that need input before/during Phase 1. Each has a proposed default so silence = proceed with the default; a redirect changes course.
+
+1. **`/customers/edd` vs `/customers/endorsements` (Bridge-style)?** Jeremy raised this. Bridge.xyz uses an endorsement-per-rail model (`base` USD, `sepa` EUR, `spei` MXN, `pix` BRL, `faster_payments` GBP, `cop`), each with its own required fields and independent lifecycle. Stripe uses a single per-customer requirements bag with country-typed fields. This doc currently uses a single `/customers/edd` bag with Lead-HNWI-shaped fields — appropriate for the US-HNWI-only Phase 1, but likely needs reshaping when SEPA / SPEI / etc. land. See the [industry-pattern comparison from the 2026-07-18 librarian analysis](https://apidocs.bridge.xyz/platform/customers/customers/endorsements) and the active Slack discussion. **Proposed default: keep `/customers/edd` for Phase 1; pivot to endorsements pattern in a follow-up PR before we ship the second compliance regime.**
+2. **`PurposeOfAccount` — need an individual variant?** The existing enum values are business-flow oriented (`CONTRACTOR_PAYOUTS`, `EMPLOYEE_PAYOUTS`, `MARKETPLACE_SELLER_PAYOUTS`, `AP_AUTOMATION`, `CROSS_BORDER_B2B`, etc.). Only `P2P_TRANSFERS`, `CHARITABLE_DONATIONS`, `OTHER` are cleanly individual-applicable. **Proposed default: add `IndividualPurposeOfAccount` alongside the business one, matching the pattern we just applied to `IndividualSourceOfFundsCategory`.**
+3. **EDD backend storage: JSON blob on `EntUmaaasIndividualUserInfo` vs a dedicated `EntUmaaasCustomerEdd` ent?** Blob is a cheaper first cut (one column, no migration for schema evolution); dedicated ent is cleaner for indexing/query/audit. **Proposed default: dedicated ent — small upfront cost, better long-term shape, matches sparkcore convention of "sandbox-adjacent state is persisted, not ephemeral" (per `grid_embedded_wallet_session_sandbox_state` and siblings).**
+
+## What exists today
+
+**API surface:**
+- `POST /customers`, `PATCH /customers/{id}` — polymorphic on `customerType` (`INDIVIDUAL` | `BUSINESS`).
+- `POST /documents` — multipart upload, attached to a `Customer` or `BeneficialOwner`.
+- `POST /verifications` — takes `{ customerId }`, returns a `Verification` with `errors[]`.
+- `POST /customers/{id}/kyc-link` — hosted-flow link (unchanged by this proposal).
+- `POST /beneficial-owners` for KYB.
+- `POST /customers/bulk/csv` — **spec-only; backend handler in sparkcore raises `NOT_IMPLEMENTED`.**
+
+**Backend behavior** (from investigation of the sparkcore handlers):
+- `POST /documents` is **synchronous and direct-to-Sumsub**. The handler does `file.read()`, creates the `EntGridDocument` row, and forwards the multipart to Sumsub in the same request. No S3, no temp buffer, no async queue.
+- Documents require a `document_holder_id` (Customer or BeneficialOwner) at creation. No orphan-doc concept in the current handler.
+- There is no TTL on documents. Persistence ends only on explicit `DELETE`.
+- Sumsub is the KYC/KYB provider. LexisNexis is used for OFAC screening on LSP switches; Striga / Bitso / Sandbox skip it.
+- PII persists in Grid's DB long-term (`EntUmaaasIndividualUserInfo`, `EntGridBusinessInformation`, `EntUmaaasUser`) — this is not proxy-only.
+
+**Sandbox surface** (existing, this proposal extends):
+- The sandbox mode uses magic-value suffixes on customer/BO fields to drive deterministic KYC/KYB outcomes without hitting real vendors. Documented in [`mintlify/snippets/sandbox-verification.mdx`](../../mintlify/snippets/sandbox-verification.mdx).
+- **Business KYB:** last 3 digits of `businessInfo.registrationNumber` → `001` PENDING / `002` REJECTED / anything else APPROVED (when `POST /verifications` runs).
+- **Beneficial-owner KYC:** last 3 chars of `personalInfo.lastName` → same 001 / 002 / else mapping.
+- Sandbox switches skip both Sumsub and LexisNexis; verification outcomes resolve entirely from the magic values.
+- **Gap:** there's no analogous magic-value pattern for individual-customer KYC via `POST /customers` today. This proposal adds one (§4 below).
+
+## Proposed changes
+
+### 1. `taxIdentification` on `IndividualCustomerFields`
+
+Today `IndividualCustomerFields` is `customerType`, `kycStatus`, `fullName`, `birthDate`, `nationality`, `address`. Add one optional field:
+
+```yaml
+taxIdentification:
+  type: object
+  required: [idType, identifier]
+  properties:
+    idType:            { $ref: ./IdentificationType.yaml }   # SSN | ITIN | EIN | NON_US_TAX_ID
+    identifier:        { type: string }
+    countryOfIssuance: { type: string, description: "ISO 3166-1 alpha-2" }
+  description: >-
+    Tax identifier. US persons: SSN or ITIN. Non-US persons: NON_US_TAX_ID with
+    countryOfIssuance. Single object; multi-jurisdictional taxpayers are a rare
+    edge case we can revisit if it becomes real demand.
+```
+
+Maps to Sumsub's `fixedInfo.tin` field internally. `IdentificationType` uses the existing enum (`SSN | ITIN | EIN | NON_US_TAX_ID`).
+
+**Photo IDs are a separate category** and stay on the `Document` resource — `POST /documents` already carries `documentNumber` alongside the file upload. Grid never accepts a driver's license or passport *in place of* an SSN for a US person; they're additive (tax ID as the identifier, photo document as identity confirmation).
+
+### 1a. Enhanced due diligence (EDD) — separate resource
+
+Enhanced-due-diligence fields don't belong on `IndividualCustomerFields` — they're only required for a narrow customer segment (HNWI, higher-risk jurisdictions, certain product tiers), they may be persisted separately on the backend (separate ent), and mixing them into every customer response bloats the base shape.
+
+Model them as a **separate linked resource**, following the "prefer flat, linked resources over deeply nested sub-resources" guidance in [`openapi/README.md`](../../openapi/README.md) (same pattern as `/customers/external-accounts?customerId=...`):
+
+```yaml
+POST /customers/edd
+body:
+  customerId: Customer:...                 # link to the parent customer
+  sourceOfFundsCategories: [...]           # $ref IndividualSourceOfFundsCategory
+  sourceOfFundsOtherDescription: <string>  # required when sourceOfFundsCategories includes OTHER
+  sourceOfWealthCategories: [...]          # $ref SourceOfWealthCategory
+  sourceOfWealthOtherDescription: <string> # required when sourceOfWealthCategories includes OTHER
+  purposeOfAccount: <enum>
+  purposeOfAccountOtherDescription: <string> # required when purposeOfAccount is OTHER
+  expectedMonthlyTransactionCount: <enum>
+  expectedMonthlyTransactionVolume: <enum>
+  annualIncomeRange:   <enum>              # e.g. UNDER_50K | RANGE_50K_100K | RANGE_100K_250K | RANGE_250K_1M | OVER_1M
+  netWorthRange:       <enum>              # e.g. UNDER_100K | RANGE_100K_500K | RANGE_500K_1M | RANGE_1M_5M | RANGE_5M_25M | OVER_25M
+  pepStatus:           <enum>              # NONE | DOMESTIC | FOREIGN | HIO | FAMILY_OR_ASSOCIATE
+response 200: <EDD payload echoed back>
+
+GET /customers/edd?customerId=Customer:...    → 200 <same shape> | 404 if never set
+PATCH /customers/edd?customerId=Customer:...  → partial update, same shape
+```
+
+Response envelope wraps just the EDD fields — nothing about the base customer changes on this resource.
+
+**How the trigger works.** When Sumsub's per-country / per-flow config or a Grid-layer overlay decides EDD is required for a specific applicant, `POST /verifications` returns `MISSING_FIELD` errors pointing at the specific EDD fields (`field: "edd.sourceOfWealthCategories"`, etc.). The platform then calls `POST /customers/edd` with the customer link + missing fields and retries `POST /verifications`.
+
+`expectedMonthlyTransactionVolume` mirrors the existing `BusinessInfo` field (single USD-equivalent bucket). Whether volume should split by fiat vs crypto is worth revisiting once we have real HNWI usage — trivially additive if needed.
+
+### 2. Machine-readable validation on 400 responses
+
+Submit endpoints (`POST /customers`, `PATCH /customers/{id}`, `POST /customers/edd`, `PATCH /customers/edd`) reject malformed field values with a **400 `INVALID_INPUT`**. When they do, the response populates `details.field` (dot-path to the offending field) and `details.constraint` (machine-readable format hint — regex, enum, min/max length) so the platform can render sensible client-side validation on its onboarding form:
+
+```json
+{
+  "status": 400,
+  "code": "INVALID_INPUT",
+  "message": "The provided tax identifier is not a valid US SSN",
+  "details": {
+    "field": "taxIdentification.identifier",
+    "constraint": {
+      "pattern": "^\\d{3}-\\d{2}-\\d{4}$",
+      "format": "us-ssn"
+    }
+  }
+}
+```
+
+`FieldConstraint.yaml` defines the `details.constraint` shape (`format`, `pattern`, `enum`, `minLength`, `maxLength`) as a shared schema so SDKs and docs get a named type.
+
+### 3. Document upload strategy
+
+**The constraint (important design context).** `POST /documents` in the current backend forwards the multipart body directly to Sumsub in the same synchronous request. Sumsub requires the `applicant_id` at doc-upload time, and Sumsub does **not** support rebinding an uploaded document to a different applicant later. Consequence: **documents cannot be uploaded before the customer exists.** This shapes what "one-shot" can mean when documents are involved.
+
+Two usage patterns for the singleton flow:
+
+**Pattern A — no-doc customers.**  
+For segments where no ID documents are required (some LSP configurations for basic domestic USD, for example), the platform calls `POST /customers` followed by `POST /verifications` — two round-trips, no docs.
+
+**Pattern B — sequential create + upload (default with docs).**  
+`POST /customers` → response gives `customer_id` → `POST /documents` (with `documentHolder: customer_id`) for each required document → `POST /verifications`. Three or more round-trips, but no new backend infrastructure. Matches how Sumsub actually works.
+
+Bulk-oriented document patterns (bulk-create-then-loop; one-shot with buffered docs and Grid-side caching) are covered in [Appendix A](#appendix-a--bulk-migration-deferred-future-work).
+
+Some platforms don't store their end-users' documents at all (they may only have identity data, not the scans). Those platforms will need to gather documents from the end-users directly during migration — which is Pattern B in a UI-driven flow.
+
+### 4. Sandbox mode
+
+Customers need to build and test their onboarding UX end-to-end before going to production, without hitting real vendors. Grid's sandbox already handles this for business KYB and beneficial-owner KYC via magic-value suffixes on customer fields; the new individual-customer KYC flow needs the same affordance so integrators can drive `PENDING` / `REJECTED` / `APPROVED` outcomes deterministically.
+
+**Extension — individual-customer KYC magic values.** For `POST /customers` with `customerType: INDIVIDUAL`, the last 3 characters of `fullName` determine the KYC-verification outcome when `POST /verifications` runs. This mirrors the existing beneficial-owner KYC pattern, plus a new `003` suffix for driving the EDD flow.
+
+| Suffix on `fullName` | Outcome on `POST /verifications` | Behavior |
+|---|---|---|
+| `001` | `verificationStatus: IN_PROGRESS`, `kycStatus: PENDING` | Verification submitted, held in review |
+| `002` | `verificationStatus: REJECTED`, `kycStatus: REJECTED` | Verification rejected |
+| `003` | `verificationStatus: RESOLVE_ERRORS` with `MISSING_FIELD` errors on `edd.*` fields | EDD required — platform calls `POST /customers/edd` (with `customerId` link + missing fields), then retries verification (with a non-`003`-suffix `fullName` on the next customer's payload, or a `fullName` PATCH, to drive APPROVED / PENDING on retry) |
+| anything else | `verificationStatus: IN_PROGRESS`, `kycStatus: APPROVED` | Verification approved |
+
+Example EDD flow:
+```
+POST /customers { customerType: INDIVIDUAL, fullName: "Jane Doe 003", ... }
+POST /verifications { customerId }
+  → verificationStatus: RESOLVE_ERRORS,
+    errors: [ { field: "edd.sourceOfWealthCategories", type: MISSING_FIELD }, ... ]
+
+POST /customers/edd
+  { customerId: "Customer:...", sourceOfWealthCategories: [SALARY],
+    purposeOfAccount: PERSONAL_SPEND, ... }
+  → 200 { EDD fields echoed back }
+
+# Platform PATCH fullName to strip the `003` (or the retry logic maps `003` to APPROVED
+# once EDD is present) and re-verify:
+POST /verifications { customerId }
+  → verificationStatus: IN_PROGRESS, kycStatus: APPROVED
+```
+
+**How each new piece behaves in sandbox:**
+
+- **`taxIdentification` (§1)** — sandbox accepts the field, persists it, but doesn't attempt any TIN-verification round-trip. No Sumsub `fixedInfo.tin` call. The value is echoed back on `GET /customers/{id}` for parity.
+- **EDD endpoint (§1a)** — sandbox accepts `POST /customers/edd`, persists the fields (per backend storage choice in [Open questions](#open-questions-for-phase-1) #3), echoes them on `GET /customers/edd?customerId=...`. Does not run any Sumsub requiredIdDocs check. Absence of EDD data plus a `fullName` suffix of `003` on the customer is what drives the `MISSING_FIELD` errors in the table above.
+- **400 `INVALID_INPUT` with `details.constraint` (§2)** — sandbox rejects malformed field values at submit time (`POST /customers`, `PATCH /customers/{id}`, `POST /customers/edd`, `PATCH /customers/edd`) with the same 400 shape production returns, including `details.field` and `details.constraint`. Integrators can wire their onboarding form's client-side validation against the sandbox before hitting production.
+
+**What sandbox does NOT do:**
+- No Sumsub calls (no applicant creation, no doc upload, no verification submission).
+- No LexisNexis / OFAC screening (sandbox switch already returns `None` from `gen_screen_customer`).
+- No real document processing — `POST /documents` succeeds and returns a fake `documentId`, no file is uploaded anywhere.
+
+**What sandbox DOES do (beyond magic-value outcomes):**
+- **Webhooks fire.** Sandbox emits `CUSTOMER.KYC_APPROVED` / `KYC_REJECTED` / `KYC_PENDING` events based on the magic-value outcome so integrators can wire and test their webhook handlers end-to-end. Same event shape as production.
+
+**Documentation.** `mintlify/snippets/sandbox-verification.mdx` covers all three verification types (business KYB, beneficial-owner KYC, individual-customer KYC — including the `003`-suffix EDD trigger). It's imported by `mintlify/api-reference/sandbox-testing.mdx` and the per-product `sandbox-testing.mdx` pages. The Mintlify docs also publish the per-jurisdiction / per-currency requirements matrix so platforms can build their onboarding form against it up-front.
+
+## Walkthrough: interactive singleton flow
+
+Platforms build their onboarding form against the documented per-jurisdiction / per-currency requirements matrix (Mintlify docs), submit what they have, and iterate on `errors[]` from `POST /verifications`.
+
+```
+End-user opens the platform's onboarding form.
+Platform renders form based on the docs. User fills what they can. Missing DOB.
+
+Platform → POST /customers
+             { customerType: INDIVIDUAL, fullName, address, nationality: BR,
+               taxIdentification: { idType: NON_US_TAX_ID, identifier, countryOfIssuance: BR } }
+        ← 201 { id, kycStatus: PENDING, ... }
+
+Platform → POST /verifications { customerId }
+        ← 200 { verificationStatus: RESOLVE_ERRORS,
+                errors: [
+                  { field: "birthDate", type: MISSING_FIELD,
+                    constraint: { format: "date" },
+                    reason: "Date of birth is required" },
+                  { type: MISSING_IDENTITY_DOCUMENT,
+                    acceptedDocumentTypes: [PASSPORT, DRIVERS_LICENSE, NATIONAL_ID],
+                    reason: "Identity document required" }
+                ] }
+
+All currently-known requirements are surfaced at once, not staged across rounds. Platform prompts the user for the missing DOB and a passport photo, then uploads both:
+
+Platform → POST /documents (multipart, documentHolder: customer_id,
+                            documentType: PASSPORT, country: BR)
+        ← 201 { id: "Document:..." }
+
+Platform → PATCH /customers/{id} { birthDate: "1990-01-15" }
+        ← 200 { customer }
+
+Platform → POST /verifications { customerId }
+        ← 200 { verificationStatus: IN_PROGRESS, errors: [] }
+
+Terminal state via webhook CUSTOMER.KYC_APPROVED / KYC_REJECTED, or by polling.
+```
+
+For an EDD-triggered customer, the second-round `POST /verifications` returns `MISSING_FIELD` errors pointing at `edd.*` fields; the platform then posts to `POST /customers/edd` (with the customerId link + requested EDD data) and retries.
+
+## Implementation phasing
+
+**Strategy: ship the API surface and sandbox behavior first so customers can integrate and test end-to-end immediately; progressively fill in real production behavior behind the same API surface.** The API contract is locked once Phase 1 ships, so production wire-up in later phases doesn't change what platforms see — it just makes their existing sandbox integrations work against real vendors.
+
+Each item below is intended to map to roughly one ticket. Sizing is rough: **S** ≈ half a day, **M** ≈ 1–2 days, **L** ≈ 3+ days.
+
+### Phase 1 — API surface + sandbox behavior (ship ASAP)
+
+**grid-api (OpenAPI spec + Mintlify docs):**
+
+- **S** — §1 `taxIdentification` (single object) addition to `IndividualCustomerFields.yaml`.
+- **S** — §2 New shared `FieldConstraint.yaml` schema (regex, format, enum, min/max length). Documented as the shape used inside `Error400.details.constraint` on submit endpoints.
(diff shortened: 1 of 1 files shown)


akanter:
* **#700** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/700?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/700?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,423 @@
...
+    description: Machine-readable validator hint so the platform can pre-validate its own form
+    properties:
+      format:    { type: string, description: "regex/format/enum name" }
+      pattern:   { type: string, description: "regex if applicable" }
+      enum:      { type: array, items: { type: string } }
+      minLength: { type: integer }
+      maxLength: { type: integer }
+```
+
+`severity: EVENTUALLY_DUE` is the important addition — it mirrors Stripe's `eventually_due` and lets a platform onboard a customer for basic activity now and progressively collect more as the customer's volume/features grow, without a `RESOLVE_ERRORS` block.
```
akanter: Not saying we shouldn't do this, but do we need this now? Also wondering if it's important to specify what we mean when we say `EVENTUALLY_DUE`. If we don't tell the customer/platform what the limitation is going to be, they don't know how important it is to supply the data or not. IOW, is this actionable and therefore useful?  How do the other platforms really use this in practice?

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,423 @@
...
+# openapi/paths/kyc/kyc_requirements.yaml (new)
+get:
+  summary: Discover the fields and documents required to KYC a customer of a given shape
+  operationId: getKycRequirements
+  tags: [KYC/KYB Verifications]
+  parameters:
+    - { name: customerType, in: query, required: true, schema: { $ref: ../../components/schemas/customers/CustomerType.yaml } }
+    - { name: region,       in: query, required: true, schema: { type: string, description: "ISO 3166-1 alpha-2 (jurisdiction)" } }
+    - { name: nationality,  in: query, required: false, schema: { type: string, description: "ISO 3166-1 alpha-2" } }
+    - { name: currencies,   in: query, required: false, schema: { type: array, items: { type: string } } }
```
akanter: Are there any other fields that might be necessary? I know that when we consider HNWI we might need additional info too. Totally possible that we could extend this API to add new fields, but I'm curious. Again, what do others in the industry do?

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,423 @@
...
+      line2:      { type: string }
+      city:       { type: string }
+      state:      { type: string }
+      postalCode: { type: string }
+  example:
+    ja-Kanji:
+      line1: 東京都渋谷区..."
+      city: 渋谷区
+      postalCode: 150-0001
+```
```
akanter: If this is _just_ for some asian countries, can/should we skip for now?

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,423 @@
...
+  type: array
+  items: { type: string }
+  description: >-
+    ISO 3166-1 alpha-2 country codes acceptable for the issuing document.
+    Present alongside acceptedDocumentTypes when the issuing country is constrained.
+```
+
+And a corresponding `issuingCountry` requirement on `POST /documents` (already there via the existing `country` field).
+
+### 6. `POST /customers/bulk` — JSON bulk migration
```
akanter: How does existing bulk CSV migration work with blobs? Could we also extend it to work with `identification_1`, `identification_2` or something to support arrays, kinda?

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,423 @@
...
+  - type: object
+    properties:
+      correlationId:      { type: string }
+      customerId:         { type: string, description: "Set if creation succeeded but verification failed." }
+      verificationErrors: { type: array, items: { $ref: ../verifications/VerificationError.yaml } }
+```
+
+This is the primary migration surface: a platform sends 1000 customers, gets back a job ID, polls until COMPLETED, then iterates the errors — each entry tells them exactly what's missing for that specific customer.
+
+### 7. One-shot atomic submit — `POST /verifications` with inline customer
```
akanter: Does this make more sense to live on a `/verifications` endpoint or `/customers` endpoint? A little confusing because what we're trying to do is create a customer, but it does need verifications? And the return type is probably more related to the `/verifications` return types.

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,423 @@
...
+
+## Compatibility with the hosted flow
+
+The hosted `POST /customers/{id}/kyc-link` flow doesn't change. A customer created via the API-first flow can still generate a KYC link if the platform wants to hand off the remaining collection to Grid. The reverse also holds: a customer who started in the hosted flow but abandoned it can be finished via API. The `region`/`nationality` on the Customer is authoritative for both.
+
+Where hosted and direct diverge is on which fields are populated. A hosted-flow customer will have everything filled in by the provider; a direct-API customer only has what the platform submitted. `GET /customers/{id}` returns the same shape regardless.
+
+## Open questions for review
+
+1. **Individual `identification[]` array vs. single object.** Real cases where a single individual needs multiple identity numbers on-file are rare (dual citizens, US residents with foreign nationality). Keep as an array from day one, or start with a single optional `identification` and upgrade later?
```
akanter: Yeah I feel like we should support an array where we can. As I mentioned, perhaps for the CSV case we can have array suffixes. Or maybe there's some other standard way of making arrays kind-of work for CSV?

What do other competitors do in this? I recognize CSV is probably the most portable, but YAML or JSON is more expressive. I also know we already have CSV so it would be nice to keep using that as much as possible

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,423 @@
...
+2. **`localizedAddress` on the wire, or only in errors.** Should JP customers submit `address.localizedAddress["ja-Kanji"]` themselves, or should Grid transliterate for them and store both scripts internally? Transliteration quality is uneven; leaning toward "platform submits both, we validate both".
```
akanter: I think/hope we can probably skip this for now? But if we wanted it I'd guess we can validate? Would validation look very different from trying to transliterate anyway?

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,423 @@
...
+3. **`EVENTUALLY_DUE` semantics.** Do we surface it now, or wait until we actually have volume-tier requirements to bind it to? Adding the enum value is cheap; over-promising the semantics is expensive.
+4. **Inline document upload size.** 5 MB base64 in a JSON body is workable but not great. Alternatives: pre-upload via `POST /documents`, get back `documentId`, then reference `documentId` in the inline body. That's an extra round-trip but handles arbitrary file sizes cleanly. Recommend supporting both.
```
akanter: Again, what's industry standard here? Do we need documents that big or can we require clients to compress first? Do we need to do anything to screen these documents first or something to ensure it's not illicit material? Do we already have an API to upload documents and return docIds? If so does it do bulk? Are we storing these documents or just passing them on? If we're passing them on, what are the requirements to our vendor(s), e.g., sumsub and lexis nexus?

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,423 @@
...
+5. **Bulk endpoint size cap.** 1000 customers/request is a starting guess. Bridge caps at 500, Circle at 100. Empirically what does a migration look like — how many customers do platforms typically bring over in one batch?
```
akanter: 500 or 1000 are both fine, we can sort that out later. I think it will depend on the customer. Some may do one at a time, others may do more.

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,423 @@
...
+6. **Do we need a separate `POST /kyc` alias?** Semantically clearer than `POST /verifications` for platforms that never do KYB. Downside: two endpoints doing the same thing. Recommend no — keep one endpoint, one mental model.
```
akanter: yeah i think one endpoint

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,423 @@
...
+7. **`identification.idType` enum coverage.** Today: `SSN | ITIN | EIN | NON_US_TAX_ID`. Need to add at least `PASSPORT`, `DRIVERS_LICENSE`, `NATIONAL_ID`, `RESIDENCE_PERMIT`. Cross-check against provider capabilities before finalizing.
```
akanter: Just to be clear, these are still just ID numbers? Some of these things can have document uploads but not all (e.g., SSN).

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,423 @@
...
+8. **Handling of `contactVerification` in the inline flow.** If a jurisdiction requires email/phone verification before KYC, the inline `POST /verifications` needs to return a `RESOLVE_ERRORS` variant that says "verify these channels first" — or we require the platform to do contact verification separately before ever calling `POST /verifications`. Recommend the former, treating it as another `VerificationError` type (e.g., `CONTACT_VERIFICATION_REQUIRED`).
```
akanter: I think this might be more of a product question, but also where does this requirement idea come from? Why would contact verification need to happen before kyc? If we can rely on the customer to just tell us the customer's phone number is verified that's certainly easiest. Or, maybe contact info falls under the `EVENTUALLY_DUE` category where we can create the account but it might be restricted until the contact info has been validated?

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,423 @@
...
+
+## Rollout
+
+- Doc-first (this PR).
+- Extensions #1 (individual fields) and #2 (VerificationError severity/code/constraint) are pure schema additions; ship them first.
+- #3 (`GET /kyc/requirements`) can ship early with a static ruleset per region, refined as we learn.
```
akanter: I think we might be defining/getting these requirements from sumsub? I think we're going to need to submit all of the data to sumsub anyway so I think that's where the requirement definitions will probably live.

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,423 @@
...
+- #6 (`POST /customers/bulk`) and #7 (inline verify) are net-new endpoints; ship them behind an opt-in header for the first few pilot integrators before making them GA.
+- No version bump required for the additive schema changes. `POST /customers/bulk` and `POST /kyc/requirements` are new operations — same major version.
```
akanter: is `/customers/bulk` entirely new?



Should we call this `/verification/requirements` instead of `kyc/requirements`?

docs/plans/2026-07-18-kyc-api-design.md
akanter: we don't have anything for EDD in this requirments I don't think. honestly a little surprising but we can go wtih it? we generally want to request as little information as possible I think? As we build, we should be able to add more required and optional fields but we should at least start with what's in this image.

docs/plans/2026-07-18-kyc-api-design.md
akanter: I actually don't see where we're supposed to even supply SSN/ITIN in the sumsub docs. Do you have a link where we're doing this? I'm trying to find in the sumsub docs so I can corroborate what the various ID types are.

docs/plans/2026-07-18-kyc-api-design.md
akanter: According to https://docs.google.com/spreadsheets/d/1pxyK7j8k865u_ZxUp1Ymjdigyo76JK7tAzsetUnP8oQ/edit?gid=1402608033#gid=1402608033 we are already collecting the phone number correctly. How are we doing it for existing customer onboarding? Is it a part of the CRB bulk CSV onboarding API?

docs/plans/2026-07-18-kyc-api-design.md
akanter: I think we should maintain a cache. We control the changes for sumsub so really we just need to refresh the cache if/when we update our flows in sumsub. And yeah you're right it doesn't change often at all. I'd guess we could have a knob to enable/disable the cache, but also just refresh the cache slowly. Probably not a problem to do it every hour or even slower (12h? daily? only on reboot?)

docs/plans/2026-07-18-kyc-api-design.md
akanter: How does our existing CSV flow work? Does it include documents at all? Are we storing the data at all? Or maybe only temporarily before we pass to sumsub? Do we also pass it to LexisNexis?

docs/plans/2026-07-18-kyc-api-design.md
akanter: Do we need taxResidencies? Where is that coming from? Does it already exist? I don't see it on the initial list of things we need.  
  
If we do need JSON, I imagine it might be `/customers/bulk/json` to mirror `/customers/bulk/csv`. But if we can keep everything in the csv flow I think that would probably be easiest for now. Semicolon delimited isn't the worst thing if we only have arrays of scalars.

docs/plans/2026-07-18-kyc-api-design.md
akanter: Like I asked elsewhere, how do we do this with the existing bulk csv endpoint for documents?

docs/plans/2026-07-18-kyc-api-design.md
akanter: Does the document upload go straight to sumsub or do we get the data first ourselves and then send it along? Is there a TTL on the data if it's not used in an entity/customer creation on a particular timeline?

docs/plans/2026-07-18-kyc-api-design.md
akanter: clearer names are great. It does seem like the existing SSN/ITIN/EIN/NON_US_TAX_ID are pretty different from the photo ID stuff. Are there circumstances where we would accept, say, a drivers license, instead of the SSN? Or would they generally be additive? It _seems_ like they are just a different category and they probably shouldn't be combined in that case, but I could be wrong?

docs/plans/2026-07-18-kyc-api-design.md
akanter: sounds good.

docs/plans/2026-07-18-kyc-api-design.md
akanter: Don't we need sourceOfFundsCategories and purposeOfAccount and expectedMonthlyTransaction* and sourceOfWealth (categories?) for the HNWI? I think the question I have here is about how we really trigger the additional requirements for HNWI? I suppose for the bulk endpoint, it should accept all the various fields but when we try to create the customer we might fail if we determine we need HNWI fields and they aren't supplied? (How would we determine that? Does sumsub tell us that?)

docs/plans/2026-07-18-kyc-api-design.md
akanter: Who added the bulk endpoint? If it's not implemented, maybe we should just go with JSON (or yaml or toml) to begin with. We should inquire about it as a design question.

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,520 @@
+# API-based KYC flow — design sketch
```
akanter: we should not call it a sketch here. Let's clean up this doc so it's nice and easy to review with the questions/design decisions clearly called out

docs/plans/2026-07-18-kyc-api-design.md
akanter: We should call it out in the design doc we have here. The plan is to send it to Victor (and others) to review. I think we should just go straight to JSON.

docs/plans/2026-07-18-kyc-api-design.md
akanter: Sounds good. FWIW, as we build out the plan for this I think while we absolutely need to keep bulk in mind I think we'll implement singleton first?

Should `sourceOfWealth` and `purposeOfAccount`also be categories/lists?  
  
Do we need to separate the volume by crypto and fiat? Or just single currency amount? (And how would we store it?)

docs/plans/2026-07-18-kyc-api-design.md
akanter: Huh. Fascinating. Sounds like either we need to cache the images (which might be expensive, might have legal consequences, though maybe less so if the data always is TTL'd?) or we need to have the doc upload be either user-driven-after-bulk or done in bulk after the initial creation? Sounds like we might really want product/customer buy-in on this? I suppose there are probably some (many?) customers who don't store the documents from their end-users, so it's very likely the customers would need to upload those documents anyway? I guess what I'm saying is we might need to support multiple scenarios, one of which includes "customers are created, but not fully available for KYC because doc upload is missing and needs end user interaction".  
  
Not sure our API agreement with sumsub, but if the docs are missing and we know they're required, we might want to skip the calls to sumsub entirely in that case.

docs/plans/2026-07-18-kyc-api-design.md
akanter: I think for the time being we should skip a self-attest flag if it's not really needed. I think these flows are only for LSP gridswitch anyway?

akanter (commented):
let's really clean up the doc and make it nice and easy to read with the questions clearly called out. Also, as far as implementation goes we'll definitely want to get the single customer flow working first and then we'll layer on bulk.  
  
Probably update the PR description too

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,520 @@
...
+
+Still deferred (not in this schema): `taxResidencies[]` (structured FATCA/CRS data), `employmentStatus`, `additionalNationalities[]`, `aliases[]`, `intermediaryStatus`, `mailingAddress`, `countryOfResidence`, `occupation`, `employer`. These are more niche EDD extras — leaving out until a concrete customer segment needs them, but the design stays additive so adding any of them later is a non-breaking change.
```
akanter: remove comment. In general, remove extraneous stuff that's not useful from the doc please.

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,520 @@
...
+  description: >-
+    ISO 3166-1 alpha-2 country codes acceptable for the issuing document.
+    Present alongside acceptedDocumentTypes when the issuing country is constrained.
+```
+
+And a corresponding `issuingCountry` requirement on `POST /documents` (already there via the existing `country` field).
+
+### 6. Bulk migration — JSON path for nested data, CSV path stays as-is
+
+Today `POST /customers/bulk/csv` handles the flat-customer case (`fullName`, `birthDate`, address components in fixed columns, plus an optional `email`). It can't represent (a) nested beneficial owners for KYB, (b) `taxIdentification[]` arrays (for dual citizens with multiple tax IDs — a plain string column works only for the single-ID case), or (c) attached documents (binary blobs — no doc columns exist today).
```
akanter: didn't we say that it's actually stubbed out? We should call that out here

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,520 @@
...
+
+**What other vendors actually do** (research summary):
+- **Stripe migration CSVs** — strict flat files, one row per child entity, multiple CSVs joined by ID column when nested data is needed ([Stripe payment-method imports](https://docs.stripe.com/get-started/data-migrations/payment-method-imports)).
+- **HubSpot** — supports separate object CSVs joined by ID, and semicolon-delimited cells (`"one;two;three"`) for simple checkbox-style multi-value fields ([HubSpot multi-checkbox](https://knowledge.hubspot.com/import-and-export/add-or-replace-multiple-checkbox-property-values)).
+- **Bridge, Alloy, Circle, Sumsub** — all skip CSV entirely for KYC data; they use JSON with arrays for nested data, and multipart uploads that return a `document_id` referenced from the customer payload.
+- **No vendor I checked uses `identification_1_type`, `identification_2_type` suffix columns.** It's a plausible workaround but not a real industry convention.
```
akanter: rm line

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,520 @@
...
+        items: { type: string }
+        description: >-
+          IDs of documents previously uploaded via POST /documents. This is the
+          only supported way to attach docs to a bulk entry — the docs must
+          exist before the bulk call, uploaded with the platformCustomerId (or
+          a temp holder ID) so they can be re-associated when the customer is
+          created.
+```
+
+**Path C. Documents.** Continue using the existing `POST /documents` endpoint (multipart, 10 MB per file per current spec). For bulk migration, the platform pre-uploads docs in a loop (or parallel), collects the returned `documentId`s, and references them by ID in the bulk-customer JSON. Rationale from the research:
```
akanter: didn't we say we can't do this without adding document caching on our end? if we're using raw sumsub customers need to be created first and then we can upload documents for them

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,520 @@
...
+- **Vendor limits are the constraint**, not ours: Sumsub caps PoA docs at 50 MB, IDs at 100 KB minimum; Bridge accepts base64 up to 24 MB. Passing through inline base64 in a bulk JSON body would balloon a 1000-customer payload past any reasonable API-gateway limit.
+- **Modern vendor pattern is upload-then-reference** (Alloy `document_token`, Sumsub `imageId`, Stripe `file.id`). Following that pattern keeps the bulk payload sane.
+- **Screening is vendor-side** — Sumsub checks authenticity, tampering, EXIF, template matching. We don't do additional malware/content screening ourselves; we pass the blob to the KYC provider and they screen it.
```
akanter: rm comment

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,520 @@
...
+      correlationId:      { type: string }
+      customerId:         { type: string, description: "Set if creation succeeded but verification failed." }
+      verificationErrors: { type: array, items: { $ref: ../verifications/VerificationError.yaml } }
+```
+
+This is the primary migration surface for anything richer than flat individuals: platform pre-uploads docs, sends 1000 customers via `POST /customers/bulk/json`, gets back a job ID, polls until COMPLETED, iterates the errors — each entry tells them exactly what's missing for that specific customer.
+
+### 7. One-shot atomic submit — `POST /customers` with `submitForVerification: true`
+
+Even with the bulk endpoint, the interactive single-customer UX still requires create → docs → verify (three round trips) to get the first "here's what's missing" response. The primary intent of that call is *creating/updating a customer*, and verification is a side effect — so the atomic mode lives on `POST /customers`, not on `POST /verifications`. This also keeps the "one endpoint, one mental model" property Aaron already flagged.
```
akanter: don't need all of the historical nitty gritty here. jeez.



anyway, same thing here - if we need to create the sumsub entity first before we can upload files then we might have an issue.

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,520 @@
...
+
+## Open questions for review
+
+Resolved from v1:
+- ~~`EVENTUALLY_DUE` semantics~~ — deferred; reframed in §2. Not adding until we have a concrete "blocks feature X at volume Y" mapping.
+- ~~`localizedAddress`~~ — deferred; won't add to wire format until we're onboarding a JP/CN/KR platform.
+- ~~One `POST /kyc` alias vs. `POST /verifications`~~ — one endpoint, per Aaron.
+- ~~Endpoint naming~~ — `/verification/requirements`, per Aaron.
+- ~~Inline atomic verify shape~~ — moved to `POST /customers` with `submitForVerification: true` flag, per Aaron.
+- ~~Inline base64 doc uploads~~ — dropped; use `documentIds` referencing pre-uploaded docs, per vendor conventions.
```
akanter: remove things here that we're already decided. The whole point of this exercise is to come up with a concise document that's easy to review. Including every single decision we made makes this more complicated

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,520 @@
...
+- ~~Split tax IDs from photo IDs~~ — done. `taxIdentification[]` on the customer for SSN/ITIN/EIN/NON_US_TAX_ID (maps to Sumsub's `fixedInfo.tin`). Photo IDs (passport, driver's license, national ID) are already carried on the `Document` resource via `documentNumber` — no schema change needed there. Mirrors Sumsub's own model.
+
+Still open:
+
+1. **NRA 2-ID rule.** The Lead Bank requirement is that non-resident aliens must provide **two** forms of ID (one documentary, one non-documentary). Now that we've split tax-IDs from photo-IDs: the rule is really "at least one `taxIdentification` entry OR at least one photo-ID document (`POST /documents` with `documentType` in `{PASSPORT, DRIVERS_LICENSE, NATIONAL_ID}`) + at least one non-doc verification". How do we express this compound rule in `GET /verification/requirements` and in the `VerificationError` shape for a violation? Leaning toward a `code: "nra_two_id_rule_not_met"` error that lists the acceptable combinations, plus a structured description in the requirements payload.
```
akanter: I'd imagine this can be represented in sumsub? Yeah it would need to indicate which document/field is missing based on the nationality, currency, etc.

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,520 @@
...
+
+2. **`contactVerification` before KYC.** Backend investigation confirms Aaron's instinct: only Striga currently requires contact verification pre-KYC (`ContactVerificationRequirements(email=True, phone=True)` in `striga_grid_switch.py`); every other switch defaults to no requirement. `POST /customers` doesn't send OTPs — it initializes verification state to PENDING only if the switch requires it, then defers to explicit `verify-email` / `verify-phone` endpoints. So this is genuinely provider-specific, not a global KYC precondition. Two remaining product questions: (a) for Striga specifically, do we want `/verification/requirements` to surface the "verify email/phone first" step so platforms can pre-drive it during onboarding rather than being surprised by a 409 on `POST /verifications`? (b) is there any appetite for a self-attestation path — the platform tells us "phone verified, we did OTP on our side" — or does Striga's compliance obligation require us to run the OTP through them? The current 409-on-`POST /verifications` behavior works either way.
```
akanter: Are we even concerned about Striga here?

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,520 @@
...
+
+3. **`POST /documents` bulk mode.** Today each doc upload is one multipart request. For migrations with many docs per customer × many customers, that's a lot of round-trips. Do we add `POST /documents/bulk` (multipart with a manifest + N files), or is per-doc parallel upload from the client good enough? Modern KYC vendors don't offer bulk doc upload endpoints (each doc is its own multipart), so this may not be a real need.
+
+4. **CSV vs JSON — do we even need CSV?** Raised by Aaron in R4. The existing `POST /customers/bulk/csv` endpoint was in the initial Grid schema commit (Jeremy Klein, Oct 2025) but has never had a backend implementation (it's a stub). Since we're implementing bulk import from scratch, do we:
```
akanter: No one cares about what revision number or that I raised it. This is "my" document so of course I raised it.

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,520 @@
...
+   - **JSON + explicit YAML/TOML** — same as JSON-only but formally supporting alternative structured formats.
+   The tradeoff is: CSV is universally understood by non-technical operators who might be doing platform migrations from spreadsheets; JSON is more expressive and better suited for anything with nested/structured data. Leaning JSON-only unless we hear that platform migration teams genuinely need CSV.
+
+5. **`documentIds` at customer creation.** When creating a customer with `documentIds`, the docs must already exist. But `POST /documents` requires `documentHolder` to be an existing Customer/BeneficialOwner ID — chicken and egg. Two options:
+   - Allow `documentHolder: "PENDING:<platformCustomerId>"` on document upload, then rebind at customer creation time.
+   - Upload docs *without* a holder, get back `documentId`s, then bind at customer creation.
+   Neither is ideal; both work. Leaning toward the platformCustomerId-based tag since it's less new state.
```
akanter: You had said we're uploading directly to sumsub. Does sumsub support rebinding document IDs? I kinda doubt it. If that's the case, we can't upload docs first.



It also makes it seem _very_ unlikelny we'd ever be able to fully create a KYC'd account in one shot then, unless we provide some doc caching ourselves before sending on to sumsub. That might be a reasonable question to bring up here

docs/plans/2026-07-18-kyc-api-design.md
akanter: You never answered this?

docs/plans/2026-07-18-kyc-api-design.md
akanter: you didn't respond here either

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,301 @@
...
+                                errors: [{ field: "birthDate", type: MISSING_FIELD,
+                                           code: "birth_date_required" }] } }
+
+User provides DOB.
+
+Platform → PATCH /customers/{id}
+             { birthDate: "1990-01-15", submitForVerification: true }
+        ← 200 { customer, verification: { verificationStatus: RESOLVE_ERRORS,
+                                          errors: [{ type: MISSING_IDENTITY_DOCUMENT,
+                                                     acceptedDocumentTypes: [PASSPORT, ...] }] } }
```
akanter: In this circumstance, since we have created the customer now shouldn't the identity document requirement be surfaced in the first round of verification failures?

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,301 @@
+# API-based KYC flow — design proposal
+
+## Context
```
akanter, line 3: we probably also want to include https://docs.google.com/spreadsheets/d/1pxyK7j8k865u_ZxUp1Ymjdigyo76JK7tAzsetUnP8oQ/edit?gid=1402608033#gid=1402608033 as a link for a reference for the KYC requirements for Lead which are addressed here

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,361 @@
...
+
+2. **Production wire-up (Phase 2 — progressive, per piece, no API changes).**  
+   Fill in real vendor behavior behind the Phase 1 API surface. Order is flexible — each item is independent:
+   - `taxIdentification[]` → Sumsub `fixedInfo.tin` mapping in `kyx_sumsub_manager.py` (so US SSN/ITIN gets sent to Sumsub for TIN-based non-doc verification).
+   - `POST /verifications` → real Sumsub applicant submission for LSP-switch customers (already exists for the current flow; extend to cover the new fields).
+   - `POST /documents` → real Sumsub doc upload (already exists; extend for any new document types).
```
akanter: Huh? I think this is already done? Or do you know we need to support more document types?

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,361 @@
...
+   - `GET /verification/requirements` → Sumsub-backed cache with 24 h TTL + invalidation endpoint + bypass knob. Phase 1 sandbox returns the real matrix; Phase 2 hooks the same code path to production.
```
akanter: I'm guessing this just wouldn't be different for sandbox vs production then? Probably not separate work.

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,361 @@
...
+
+## Implementation phasing
+
+**Strategy: ship the API surface and sandbox behavior first so customers can integrate and test end-to-end immediately; progressively fill in real production behavior behind the same API surface.** The API contract is locked once Phase 1 ships, so production wire-up in later phases doesn't change what platforms see — it just makes their existing sandbox integrations work against real vendors.
+
+1. **API surface + sandbox behavior (Phase 1 — ship ASAP).**  
+   Singleton flow only. All the schema additions and new endpoint(s) from the singleton scope: §1 (`taxIdentification`), §1a (optional EDD extras), §2 (`VerificationError` code + constraint), §3 (`GET /verification/requirements`), §5 (`submitForVerification` flag), and §7 sandbox magic values for individual-customer KYC (last-3-chars of `fullName` → `001` PENDING / `002` REJECTED / else APPROVED). All schema-locked; the sandbox switch resolves outcomes deterministically without calling Sumsub or LexisNexis. Customers can integrate their full onboarding UX against this in sandbox before any production behavior is in place.
```
akanter: Can you break out the required changes into their own bullets that will roughly map to tickets (and you could size them)

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,382 @@
...
+- **S** — Mintlify — extend `sandbox-kyb-verification.mdx` with an "Individual customer verification" subsection (fullName-suffix outcome table).
+
+**webdev/sparkcore (backend):**
+
+- **S** — Accept + persist `taxIdentification[]` on `POST /customers` (INDIVIDUAL); JSON blob on `EntUmaaasIndividualUserInfo`.
+- **S** — Accept + persist EDD-extras fields (§1a) on `POST /customers`.
+- **S** — `submitForVerification` handling in `POST /customers` / `PATCH /customers/{id}`; atomic response wrapper.
+- **S** — Populate `VerificationError.code` + `.constraint` in existing error paths (Phase 1 uses deterministic Grid-side codes; Phase 2 swaps in real Sumsub reasons behind the same field).
+- **M** — Sandbox switch — individual-customer KYC magic values (`fullName` suffix → `001` PENDING / `002` REJECTED / else APPROVED on `POST /verifications`).
+- **S** — Sandbox switch — mock `POST /documents` returns a fake `documentId` (no bytes uploaded).
```
akanter: Does this not exist already?

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,382 @@
...
+### Phase 1 — API surface + sandbox behavior (ship ASAP)
+
+**grid-api (OpenAPI spec + Mintlify docs):**
+
+- **S** — §1 `taxIdentification[]` addition to `IndividualCustomerFields.yaml`.
+- **S** — §1a Optional EDD-extras additions (`sourceOfWealthCategories`, `sourceOfFundsCategories`, `purposeOfAccount`, `expectedMonthlyTransactionCount/Volume`, `annualIncomeRange`, `netWorthRange`, `pepStatus`).
+- **S** — §2 `VerificationError.yaml` — add optional `code` + `constraint` fields.
+- **M** — §3 `GET /verification/requirements` endpoint spec + `VerificationRequirements.yaml` response schema.
+- **S** — §5 `submitForVerification` boolean flag on `POST /customers` / `PATCH /customers/{id}`; wrapped `{customer, verification}` response schema.
+- **S** — Mintlify — extend `sandbox-kyb-verification.mdx` with an "Individual customer verification" subsection (fullName-suffix outcome table).
```
akanter: should this be a separate document or should we rename it to include kyc info?

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,382 @@
...
+- **M** — `taxIdentification[]` → Sumsub `fixedInfo.tin` mapping in `kyx_sumsub_manager.py` (US SSN/ITIN sent to Sumsub for TIN-based non-doc verification).
+- **M** — Swap the Phase-1 deterministic `VerificationError.code` / `.constraint` values for real Sumsub-returned reasons on LSP-switch verification paths.
+
+### Phase 3 — Bulk migration
+
+§4 `POST /customers/bulk/json` once the singleton flow is validated in production. Requires a job-execution backend that doesn't exist today.
+
+### Phase 4 — Document caching / Pattern D
+
+Only if product & compliance sign off. Independent of everything above. Enables one-shot bulk migration with documents; blocked by legal + infra + cost decisions per §6.
```
akanter: I think we can probably just say that these would be worked on later. Not as a part of this workstream.

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,381 @@
...
+**Pattern A — no-doc customers (true single-request).**  
+For segments where no ID documents are required (some LSP configurations for basic domestic USD, for example), a single `POST /customers` with `submitForVerification=true` creates and submits atomically. Nothing else needed.
+
+**Pattern B — sequential create + upload (default with docs).**  
+`POST /customers` → response gives `customer_id` → `POST /documents` (with `documentHolder: customer_id`) for each required document → `POST /verifications` (or `submitForVerification=true` on the final PATCH). Two or more round-trips, but no new backend infrastructure. Matches how Sumsub actually works.
+
+**Pattern C — bulk create then loop uploads.**  
+`POST /customers/bulk/json` creates N customers → response has each customer's `id` → platform uploads docs per-customer via `POST /documents` in a loop (or in parallel). Works with the existing backend today.
+
+**Pattern D — one-shot with docs (requires new capability, not proposed for initial implementation).**
```
akanter: these are future work because it's for bulk (which we said is future work).

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,381 @@
...
+
+Platform → POST /documents (documentHolder: customer_id, multipart)   # per customer, per doc
+Platform → POST /verifications { customerId }                          # per customer
+```
+
+## Design decisions still open
+
+1. **Pattern D — Grid-side document caching for one-shot flows with documents.** Do we build this? Real product / infra / compliance question. Default answer if we don't: sequential upload (Pattern B / C).
+
+2. **NRA 2-ID rule expression.** For non-resident aliens, one tax ID + one photo ID document is required. How do we express this compound rule in `VerificationError`? Leaning toward a dedicated `code: "nra_two_id_rule_not_met"` with acceptable combinations enumerated in the `reason`, plus a structured description in the requirements discovery response. Sumsub likely surfaces this already; we can pass the requirement through as-is.
```
akanter: why is this even really question? Tax ID is a separate field from a photo id. It should just list the fields that are missing given this particular customer?

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,381 @@
...
+
+For customers that also need documents, the platform loops after the bulk create:
+
+Platform → POST /documents (documentHolder: customer_id, multipart)   # per customer, per doc
+Platform → POST /verifications { customerId }                          # per customer
+```
+
+## Design decisions still open
+
+1. **Pattern D — Grid-side document caching for one-shot flows with documents.** Do we build this? Real product / infra / compliance question. Default answer if we don't: sequential upload (Pattern B / C).
```
akanter: most of these questions are bulk-only so we should not really be prioritizing them

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,382 @@
...
+
+For customers that also need documents, the platform loops after the bulk create:
+
+Platform → POST /documents (documentHolder: customer_id, multipart)   # per customer, per doc
+Platform → POST /verifications { customerId }                          # per customer
+```
+
+## Design decisions still open
+
+Nothing blocking for the singleton flow. Tax IDs and photo IDs are separate categories in the schema (§1 `taxIdentification[]` vs. `POST /documents` `documentNumber` + `documentType`), so the requirements discovery endpoint and `VerificationError` naturally list whichever fields/documents are missing for a given customer — no compound-rule design work needed.
```
akanter: If there's nothing here, then why is this line here!? Please go through the document and make sure to edit it so that it all makes sense.

docs/plans/2026-07-18-kyc-api-design.md
```diff
@@ -0,0 +1,352 @@
...
+PATCH /customers/edd?customerId=Customer:...  → partial update, same shape
+```
+
+Response envelope wraps just the EDD fields — nothing about the base customer changes on this resource. Backend storage is TBD: a JSON blob on `EntUmaaasIndividualUserInfo` is the low-cost option, a separate `EntUmaaasCustomerEdd` ent is the cleaner-long-term option (better for indexing / query / audit). Ship whichever is faster; migrate later if needed.
+
+**How the trigger works.** When Sumsub's per-country / per-flow config or a Grid-layer overlay decides EDD is required for a specific applicant, `POST /verifications` returns `MISSING_FIELD` errors pointing at the specific EDD fields (`field: "edd.sourceOfWealthCategories"`, etc.). The platform then calls `POST /customers/edd` with the customer link + missing fields and retries `POST /verifications`.
+
+`expectedMonthlyTransactionVolume` mirrors the existing `BusinessInfo` field (single USD-equivalent bucket). Whether volume should split by fiat vs crypto is worth revisiting once we have real HNWI usage — trivially additive if needed.
+
+### 2. Machine-readable validation on 400 responses (not on `VerificationError`)
```
akanter: WHAT'S WITH THIS "(not on `VerificationError`)"?!? How many times have I told you to make things sensible as though you're reading them for the first time without previous revisions of the code or the document?! for crying out LOUD. If you want to let me, the reviewer, know something, then put up a PR comment not a code/documentation comment!! JEEEZ

PR #736, opened 2026-07-23
feat: add Bitcoin L1 deposit-address payment instruction

peterrojs:
## Summary

Adds the public API surface for a Bitcoin L1 crypto on/off-ramp on the quote/execute flow: a new `BITCOIN_L1` payment instruction. Additive and optional, so no version bump.

## Changes

- **New `PaymentBitcoinDepositAddressInfo` payment instruction** — a member of the `PaymentInstructions.accountOrWalletInfo` oneOf with a new `BITCOIN_L1` discriminator value. Carries the on-chain deposit `address` and an optional `network` (fixed to `BITCOIN`). Modeled exactly like its cousins: it inherits `BasePaymentAccountInfo` and narrows the shared `accountType` to its discriminator value, with `BITCOIN_L1` added to the shared `PaymentAccountType` enum.
- `BITCOIN` is already a valid `cryptoNetwork` for `RealtimeFundingQuoteSource` (currency=BTC, cryptoNetwork=BITCOIN); no change needed.
- **CI:** bounded the oasdiff breaking-changes PR comment so it fits GitHub's comment-size limit, with the full report in the job summary and an artifact.

## Deferred: `refundAddress`

Earlier revisions added an optional `refundAddress` to both `AccountDestination` and `RealtimeFundingQuoteSource`. Both are now removed and deferred to a follow-up:

- **`AccountDestination`** ([thread](https://github.com/lightsparkdev/grid-api/pull/736#discussion_r3670181155)) — a failed crypto payout returns funds to a Grid-controlled address that's already known, so the caller has nothing to supply. The field would also have surfaced on fiat account destinations, where a refund *address* is meaningless.
- **`RealtimeFundingQuoteSource`** ([thread](https://github.com/lightsparkdev/grid-api/pull/736#discussion_r3686228352)) — a raw address bypasses external-account whitelisting and approval (`ExternalAccount.status`, plus the trust/untrust flow). Referencing an external account is the better model, but `ExternalAccountType` has no Bitcoin L1 type yet, so that needs its own change.

Deferring costs nothing today — nothing consumes these fields yet — and re-adding either is additive/non-breaking, whereas shipping and then removing would be breaking.

## Notes

- Unblocks the server-side Bitcoin L1 quote/execute ramp, which currently uses stopgaps for these types.
- `PaymentAccountType` is referenced only via `BasePaymentAccountInfo`, so the new `BITCOIN_L1` value surfaces only on payment-instruction members.
- **Carries genuinely breaking response changes** and gets the `breaking-change` label. The errors are `response-property-one-of-added` (the new oneOf member appearing in quote and transaction responses); the warnings are the new `BITCOIN_L1` enum value propagating to the `accountType` of every member that inherits `BasePaymentAccountInfo`. All additive — no property or enum-value removals — so this is the expected shape for a new payment-instruction member, gated by API-reviewer approval per the repo's workflow design.
- SDKs (grid-typescript, grid-kotlin) are regenerated by the Stainless GitHub Action from the bundled spec + config; no SDK code is hand-edited in this PR.

## Test plan

- `npm run lint:openapi` (redocly bundle + redocly lint + spectral lint): passes with 0 errors.
- Confirmed no `refundAddress` remains in `openapi/`, `openapi.yaml`, or `mintlify/openapi.yaml`.
- After the `main` merge, re-ran `make build` and confirmed the committed bundles are byte-identical to a fresh build (no drift).
- ⚠️ Bolt pushes do not trigger this repo's `pull_request` OpenAPI workflows, so the checks on the latest commits have not re-run — see the note in the comments.

Original PR: https://github.com/lightsparkdev/grid-api/pull/732


akanter:
* **#736** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/736?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/736?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

openapi/components/schemas/common/PaymentBitcoinDepositAddressInfo.yaml
```diff
@@ -0,0 +1,21 @@
+title: Bitcoin L1 Deposit Address
```
akanter, line 1: Should this be using BasePaymentAccountInfo.yaml ?

openapi/components/schemas/common/PaymentBitcoinDepositAddressInfo.yaml
```diff
@@ -0,0 +1,22 @@
...
+required:
+  - accountType
+  - address
+properties:
+  accountType:
+    type: string
+    description: Type of payment account or wallet
+    enum:
+      - BITCOIN_L1
+    example: BITCOIN_L1
```
akanter: I think we don't need this if we're using PaymentAccountType.yaml? and I think we should be doing that based on my read of the cousin models

openapi/components/schemas/common/PaymentBitcoinDepositAddressInfo.yaml
akanter: I'd think we should probably add Bitcoin_L1 as a PaymentAccountType, and then restrict this new `PaymentBitcoinDepositAddressInfo.yaml` to use Bitcoin_L1. I don't see a reason this network/wallet/payment address should be any different from its cousins.

peterrojs: done

akanter (commented):
lgtm, but i'm guessing @shreyav or @jklein24 should take a look here :pray:

openapi/components/schemas/quotes/AccountDestination.yaml
```diff
@@ -21,4 +21,10 @@ allOf:
               The payment rail to use for the transfer. Must be one of the rails
               supported by the destination account. If not specified, the system
               will select a default rail.
+      refundAddress:
+        type: string
+        description: >-
+          Optional address the funds are returned to if the payout cannot be
+          completed. Applies to crypto off-ramp destinations.
+        example: bc1qar0srrr7xfkvy5l643lydnw9re59gtzzwf5mdq
```
akanter: The PR description is a little confusing - it says this is for offramps? For offramp refunds (e.g., USDB -> BTC offramp fails, so we want to refund) shouldn't that automatically just go back to the source, and would we ever not know that value?  
  
I think it makes sense to me that this is not the case for onramps - if the flow fails I'm guessing you should be able to specify where your BTC should go back to if it doesn't work.

jklein24: Yeah I'm slightly confused with the placement here as well. Why is this inside of the destination object?

openapi/components/schemas/quotes/RealtimeFundingQuoteSource.yaml
```diff
@@ -31,6 +31,13 @@ allOf:
           on, so the correct deposit address can be generated. Example values:
           `SOLANA`, `ETHEREUM`, `BASE`, `POLYGON`, `SPARK`, `LIGHTNING`, `BITCOIN`.
         example: SOLANA
+      refundAddress:
```
jklein24: I'm wondering if it might be good to use an external account here for whitelisting and compliance reasons. We have a status field on external accounts that lets us approve things ahead of time, etc. We also may add more there (see the slack thread I sent you)

peterrojs: let's do b)

.github/workflows/openapi-breaking-changes.yml
jklein24, line 1: Thanks for fixing this :-)

jklein24 (approved):

PR #730, opened 2026-07-22
feat(grid-wallet-prod): fork the wallet demo into a production demo (phone + API panel, passkey-only)

pengying:
## Summary

Forks `components/grid-wallet-demo` → `components/grid-wallet-prod` per the spec in the parent PR:

- **Fork**: byte-faithful copy; package renamed, dev/start on port **4001** (demo keeps 4000), `dev:apple` dropped, new README, HANDOFF.md removed. The demo and its docs embed are untouched.
- **Sidebar removed**: `page.tsx` rewritten to phone + resizable API panel only; `ConfigurePanel` deleted; mobile configure⇄playground chrome (back pill, explore CTA, blur tray) removed from TSX and SCSS.
- **Hardcoded flow**: `useWalletDemoLogic` pins `persona: 'fintech'` and `methods: ['passkey']`; `setPersona`/`useCase`/`setUseCase`/`toggleMethod` pruned. The phone's sign-in offers only Continue with Passkey.
- **Fix**: `useColumnResize` no longer reserves the removed 400px configure column, so the API panel can use the full width (verified by live drag test: cap 880px → 1280px at a 1600px viewport).

Deferred to Phase 2 (live API integration spec): embed-path default-width math, orphaned `UseCasePicker`/`USE_CASES` pruning, `layout.tsx` metadata copy.

## Verification

`npm run build` + `tsc` clean at every step; Playwright smoke test on :4001 confirmed no sidebar, passkey-only sign-in, and the API panel appending calls as the phone flow runs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #729, opened 2026-07-22
docs: design spec + plan for grid-wallet-prod (phase 1: fork, remove config sidebar)

pengying:
## Summary

Design spec and implementation plan for forking `components/grid-wallet-demo` into `components/grid-wallet-prod`, the starting point for a live production demo (real Grid API requests in Phase 2).

Phase 1 scope (implemented in the stacked PR above): fork the app to port 4001, remove the `ConfigurePanel` sidebar so the layout is phone + API panel, hardcode passkey sign-in with the Fintech use case. Live production API integration is explicitly deferred to a Phase 2 spec.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #740, opened 2026-07-24
feat(grid-wallet-prod): live Grid API — real passkey auth, funding, and signed cash-out (Phase 2)

pengying:
## Summary

Wires `components/grid-wallet-prod` from fully-scripted to the **real Grid API** (public sandbox), per the Phase 2 spec (`docs/superpowers/specs/2026-07-22-grid-wallet-prod-live-api-design.md`). The API panel now renders the requests/responses that actually went over the wire.

**The full money loop is live-verified end to end:** passkey sign-in → Add $20 (real platform on-ramp, `COMPLETED`) → Cash out $5 (quote → browser-signed `Grid-Wallet-Signature` execute → polled to `COMPLETED` in ~62s), with the displayed balance tracking the real account throughout.

### What's in here

- **Server proxy** (`/api/grid/[...path]`): allow-listed paths only, Basic-auth injected server-side, `{request,response}` envelopes with redacted auth — the panel logs real traffic.
- **Browser crypto** (`gridCrypto.ts`): TEK keygen, HPKE encrypt/decrypt, Turnkey stamps via `@turnkey/*` — session keys never leave the client; cross-checked against `scripts/embedded-wallet-sign.js`.
- **Real auth** (`gridSession.ts`): first-run EMAIL_OTP bootstrap (magic `000000` in sandbox) → registers a real Grid `PASSKEY` credential via WebAuthn → returning sign-ins are pure passkey (challenge → `.get()` → verify).
- **Webhook receiver** (`/api/webhooks`): `X-Grid-Signature` P-256 verification against `GRID_WEBHOOK_PUBKEY`, event ring buffer + poll route.
- **Live money**: balance/activity reads, Add money via the real platform on-ramp (the customer-wallet sandbox fund endpoint mints unspendable book balance — discovered live), outbound quote/stamp/execute/poll with completion checkmarks gated on `COMPLETED`.
- 38 vitest tests across the pure modules; every task browser-verified against the live sandbox.

### Env contract

`GRID_CLIENT_ID/SECRET`, `GRID_API_BASE_URL`, `GRID_CUSTOMER_ID`, `GRID_WEBHOOK_PUBKEY` in `.env.local` (git-ignored). Prod cutover = swap those values; first tap re-runs passkey registration against the production customer.

### Tracked pre-prod follow-ups (not in this PR)

- ~~Gate `GET /api/webhooks/events`~~ — route removed; deliveries are pushed over SSE instead of polled.
- Proxy customer-scoping before production keys (today it's a single-customer sandbox gateway).
- ~~Wire real activity rows onto the phone~~ — done below.
- Phone-side failure states for execute errors (panel-only today).
- Report upstream: `POST /quotes` returns undocumented 200 for platform-sourced quotes.

---

## Update — funding flows, webhook-driven arrivals, sandbox gating

A second pass over the same component. Everything below was verified against the live sandbox in a browser; the API panel now shows **only** real traffic.

### Auth follows the documented bootstrap order

- Sign-in leads with the credential a Global Account is actually born with (`EMAIL_OTP`), with the entry step prefilled from the credential's own `nickname` — no invented address.
- A passkey is added **later**, from the wallet, as its own signed action: one device ceremony authorized by the live session (it used to register *and* immediately authenticate, so a single tap asked for approval twice).
- The CTA advertises the credential `signIn` will really use. Passkeys are device-bound, so a passkey listed on the account but registered elsewhere no longer sends this browser into an assertion it can't satisfy.
- A lapsed session (15 min) can now re-authenticate in place: the OTP sheet has a surface on the wallet screen, and the WebAuthn call runs inside the tap's user activation instead of behind the re-auth round-trip.

### Funding uses the endpoints Grid actually supports

- **Add money → country → deposit instructions**: the customer's own account/routing/rails/reference, read from `GET /customers/internal-accounts`. (The EUR block is a clearly-marked placeholder IBAN in `data/placeholderDeposit.ts` until the customer has an EUR account — the read already returns one section per fiat account, so a real one appears with no code change.)
- **…or "add an account"** to pull from. A pull quote sources the **ExternalAccount** — the previous body pointed at the platform's internal account. Grid rejects `/execute` for such a quote (*"funds must be pushed … use POST /sandbox/send"*), so the quote is created and left pending, and `POST /sandbox/send` stands in for the push. Amounts here are USD cents, not USDB micro-units.
- **Add from crypto**: Base / Solana / Ethereum, network → amount → the real Grid-provisioned deposit address for that chain.
- Saved banks are seeded from `GET /customers/external-accounts` (filtered to USD/EUR, ACTIVE); selecting one quotes against that account instead of re-creating it — which also fixes the stale-recipient gap noted in #740's original follow-ups.

### Webhooks

- **`X-Grid-Signature` parsing was rejecting every real delivery.** Grid sends `{"v":"1","s":"…"}` (or a bare base64 header); we required `{"version":1,"signature":"…"}`, so verification never ran and everything 401'd. The earlier tests signed using our own assumed shape, so they passed — they now assert the documented formats.
- Verified deliveries **stream to the panel over SSE** (`/api/webhooks/stream`); nothing polls. The bus is pinned to `globalThis` because Next gives each route handler its own module instance — the publisher and subscriber were otherwise in different registries.
- **An arrival, not a tap, is what says money landed**: `INCOMING_PAYMENT.COMPLETED` drives the toast and Activity row from its payload and re-reads the balance from the API. Confirming an add no longer claims anything.

### Panel, wallet, packaging

- The panel shows only real requests: every synthesized call and response body is gone (including a fabricated `POST https://your-app.com/webhooks/grid` and ~330 lines of stubbed quote/transaction/session responses). Flows with no client call behind them log nothing.
- Activity renders the real `GET /transactions` history (EXPIRED quotes filtered out); the balance hero shows **available** with the account **total** beneath, since sandbox book balance runs ahead of spendable.
- Bank picker trimmed to the US + euro area (ACH + SEPA); APY 3%.
- `Dockerfile` + `.dockerignore` on `output: 'standalone'`. **`NEXT_PUBLIC_GRID_SANDBOX` is a build arg** (NEXT_PUBLIC_* is inlined at build time) and gates the sandbox-only "simulate funding" affordance — a production image must be built without it.
- `scripts/ensure-business-customer.mjs`: idempotent `BUSINESS` customer provisioning, writes `GRID_BUSINESS_CUSTOMER_ID` to `.env.local`.

60 vitest tests (was 38).

### Still open

- A received webhook only drives the *add* path; `INCOMING_PAYMENT.PENDING` and `OUTGOING_PAYMENT.*` are logged but don't touch the wallet.
- The sandbox funding stand-in for the *instructions* screen still runs the platform on-ramp (no quote exists there to `/sandbox/send`).
- USDB-native funding over Spark works (`REALTIME_FUNDING` + `cryptoNetwork: SPARK`, rate 1.0, verified) but isn't offered in the UI — both current paths convert.
- `docker build` unverified: the daemon wasn't running here. The standalone server itself was run and exercised (page, webhook, SSE, proxy → Grid all responding).


pengying:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/740?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#740** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/740?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/740?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#730** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/730?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#742](https://github.com/lightsparkdev/grid-api/pull/742) <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/742?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#729** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/729?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

mariano-macri (approved):

PR #742, opened 2026-07-25
fix(grid-wallet-prod): replay missed webhooks to reconnecting SSE clients

pengying:
## Summary

Fan-out to multiple clients already worked; **replay on reconnect didn't**. This adds it.

The demo drives one hardcoded customer, so every connected panel is watching the same account and should see every delivery. There is no per-client filtering — `pushEvent` walks all subscribers — and that part is verified below. The gap was that a client only ever saw events that arrived *while it was connected*, so an `EventSource` auto-reconnect after a network blip silently dropped whatever landed in the gap.

### What changed

- Every frame now carries an `id:` (a per-process sequence number). `EventSource` echoes the last one back as `Last-Event-ID` when it reconnects, and the stream replays events after that point.
- **A fresh client replays nothing.** It presents no id, and replaying history would render old deliveries as if they had just arrived — the panel timestamps entries on receipt.
- **Subscribe happens before replay**, so a delivery landing mid-replay queues behind it rather than slipping between the two steps.
- The connect handshake comment reports `clients=N`. `EventSource` ignores comment frames; it shows up in `curl -N .../api/webhooks/stream` when you want to confirm every panel is attached.

### Verified

Three concurrent clients, one delivery each time, with a disconnect and a reconnect in the middle:

```
A handshake: : connected clients=1     B: clients=2     C: clients=3
deliver #1 → A [1] | B [1] | C [1]                    ← every client, no filtering

B disconnects; #2 and #3 land while it is away
  A [1,2,3] | C [1,2,3]                               ← survivors unaffected
B reconnects with Last-Event-ID: 1
  B replayed: [2,3]                                   ← exactly what it missed
D connects fresh: []                                  ← no invented history
deliver #4 → A [1,2,3,4] | C [1,2,3,4] | B [2,3,4] | D [4]
```

Signature verification is untouched; the replay path only re-sends events that already passed it.

### Limits (documented in the module)

- The bus is **per server process**. With multiple replicas, only the instance that received a delivery can fan it out to its own clients — a shared channel (Redis pub/sub) would be needed.
- The ring buffer holds the **last 50** events, so a client away longer than that loses the overflow.

Stacked on #730.

### Update: catch-up must not duplicate either

Self-review caught a flaw in the first push. Subscribe-then-replay avoids gaps but **double-sends**: an event landing between `subscribe` and the replay loop goes out live *and* again from the backlog. Replay-then-subscribe has the opposite bug (it loses that event). The stream now subscribes immediately but **holds** live events, replays the backlog, then flushes the held ones the replay didn't cover.

Verified by reconnecting from `Last-Event-ID: 1` while deliveries stream in:

```
client saw: 2,3,4,5,6
duplicates: none
gaps      : none (2..6 as expected)
```

Also added `webhookEvents.test.ts` covering the bus contract: fan-out to every subscriber, survival of a subscriber that throws, the `seq`/`eventsSince` replay windows, and the ring dropping the overflow a long-gone client can no longer catch up on. **69 tests** total (was 60).

### Update: the stream was readable by any site

Second self-review finding, and a more serious one. `/api/webhooks/stream` is unauthenticated and carries whatever Grid delivered — transaction ids, amounts, and `counterpartyInformation`, which per Grid's schema can include a counterparty's **name, birth date and nationality**. That is the same exposure `GET /api/webhooks/events` was deleted for earlier in this stack; the SSE replacement inherited it.

An `EventSource` opened by another site is now refused (`Sec-Fetch-Site: cross-site` → 403), so a page on someone else's origin can't attach to a panel running on your machine:

```
same-origin (our own panel)        → 200
no header (curl / scripts)         → 200
cross-site (another website)       → 403
```

**This is not access control**, and the code comment says so: anyone who can reach the host can still read the stream, which behind an ngrok tunnel means anyone with the URL. Gating it properly — a session cookie, or not exposing it past localhost — belongs on the pre-prod list next to proxy customer-scoping.

https://claude.ai/code/session_01Pjqydq34z8DBK1kBrioXK2

### DIFF
--- a/components/grid-wallet-prod/src/app/api/webhooks/stream/route.ts
+++ b/components/grid-wallet-prod/src/app/api/webhooks/stream/route.ts
@@ -18,7 +18,28 @@ const KEEPALIVE_MS = 15_000;
  * one back as `Last-Event-ID` when it reconnects, and anything that landed during
  * the gap is replayed — so a blip doesn't silently drop deliveries.
  */
+/**
+ * Refuse a subscription initiated by ANOTHER site. Browsers set Sec-Fetch-Site on
+ * EventSource requests, so a page on evil.example can't quietly attach to a panel
+ * running on your machine. It is deliberately narrow: only an explicit
+ * `cross-site` is rejected, so `curl -N` and the verification scripts (which send
+ * no such header) still work.
+ *
+ * This is NOT access control. The stream carries whatever Grid delivered —
+ * transaction ids, amounts, and `counterpartyInformation`, which per Grid's schema
+ * can include a counterparty's name, birth date and nationality — and anyone who
+ * can reach the host can read it. Behind a tunnel, that's anyone with the URL.
+ * Gating it properly (a session cookie, or not exposing it beyond localhost) is a
+ * pre-prod requirement, tracked alongside proxy customer-scoping.
+ */
+function crossSite(req: Request): boolean {
+  return req.headers.get('sec-fetch-site') === 'cross-site';
+}
+
 export async function GET(req: Request) {
+  if (crossSite(req)) {
+    return new Response('cross-site subscription refused', { status: 403 });
+  }
   const encoder = new TextEncoder();
   let unsubscribe: (() => void) | undefined;
   let keepalive: ReturnType<typeof setInterval> | undefined;
--- a/components/grid-wallet-prod/src/app/api/webhooks/stream/route.ts
+++ b/components/grid-wallet-prod/src/app/api/webhooks/stream/route.ts
@@ -1,4 +1,4 @@
-import { subscribe, type WebhookEvent } from '@/lib/webhookEvents';
+import { eventsSince, listenerCount, subscribe, type WebhookEvent } from '@/lib/webhookEvents';
 
 export const runtime = 'nodejs';
 export const dynamic = 'force-dynamic';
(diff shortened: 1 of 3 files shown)


pengying:
* **#745** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/745?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#744** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/744?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#742** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/742?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/742?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#730** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/730?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#740](https://github.com/lightsparkdev/grid-api/pull/740) <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/740?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#729** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/729?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

pengying:
@greptileai review

Context for the review: this is a small SSE change with two self-review fixes already applied — reconnect catch-up (hold-and-flush so it neither gaps nor duplicates) and a cross-site subscription refusal. The parts I'd most like a second pair of eyes on:

1. `src/app/api/webhooks/stream/route.ts` — the hold/replay/flush ordering, and cleanup on abort (unsubscribe + clearInterval + controller.close).
2. `src/lib/webhookEvents.ts` — the bus is pinned to `globalThis` because Next gives each route handler its own module instance; is that the right call here?
3. Whether refusing only `Sec-Fetch-Site: cross-site` is the right narrowness, given the stream is otherwise unauthenticated (documented as a pre-prod gap).

mariano-macri (dismissed):

pengying:
### Merge activity

* **Jul 27, 9:10 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/lightsparkdev/grid-api/742).
* **Jul 27, 9:10 PM UTC**: [Graphite](https://app.graphite.com/github/pr/lightsparkdev/grid-api/742) couldn't merge this pull request because a downstack PR #729 failed to merge.
* **Jul 27, 10:01 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/lightsparkdev/grid-api/742).
* **Jul 27, 10:01 PM UTC**: @pengying merged this pull request with [Graphite](https://app.graphite.com/github/pr/lightsparkdev/grid-api/742).

mariano-macri (approved):

PR #746, opened 2026-07-26
fix(grid-wallet-prod): npm ci --ignore-scripts so the image can build

mariano-macri:
## The Dockerfile added in #740 has never produced an image

`npm ci` dies in the deps stage. The seven `@central-icons-react` packages each run a postinstall `license-check.js` that hard-fails unless `CENTRAL_LICENSE_KEY` is set, and that variable appears nowhere in this repo:

```
npm error path /app/node_modules/@central-icons-react/round-filled-radius-2-stroke-2
npm error command sh -c node ./license-check.js
npm error Error: Central Icons license key is not set.
          Please set CENTRAL_LICENSE_KEY in your environment
```

#740 noted the build as *unverified* ("the daemon wasn't running here"). It's not merely unverified — it cannot succeed as written.

## Why `--ignore-scripts` is the right fix here

`components/grid-wallet-prod/vercel.json` already sets:

```json
{ "installCommand": "npm install --ignore-scripts" }
```

That's why the Vercel deployments of this component work and the container build never did. This aligns the Dockerfile with the install behavior the component is actually developed and shipped against. No dependency here needs a postinstall to produce a working build.

## Verified end to end (linux/arm64)

- image builds, 332MB
- `/app/server.js` is the entrypoint (flat, not nested — this component has its own lockfile and the repo root has no `workspaces`)
- `NEXT_PUBLIC_GRID_SANDBOX` inlines into the client bundle
- the server boots and serves `/` with **200** carrying no Grid credentials
- an unsigned `POST /api/webhooks` is rejected **401** by the signature check

## Scope

Dockerfile only. Splitting this out from the CI-workflow PR that stacks on top of it, since this is a bug fix you want whether or not the tooling-cluster deployment ever happens.


mariano-macri:
@greptileai review

No new commits since your last review — re-triggering for a fresh pass alongside the rest of the stack.

mariano-macri:
@greptileai review

Rebased onto `07-24-fix_...replay_missed_webhooks` (#742) so this PR no longer sits on top of a draft — commit is now `5b58855`, content unchanged (1 file, +8/-1).

pengying (dismissed):

pengying (approved):

PR #747, opened 2026-07-26
ci: build and push grid-wallet-prod image to ECR

mariano-macri:
Stacked on #746 (which makes the image buildable at all).

Builds `components/grid-wallet-prod` and pushes it to the ECR repo vended by `tooling-infra` (`terraform/apps/grid-wallet-prod`), so the app can be deployed to the internal tooling EKS cluster behind Okta.

## What it does

On pushes to `main` touching `components/grid-wallet-prod/**`: assume `github-actions-grid-wallet-prod` via GitHub OIDC, build `linux/arm64`, push an immutable `sha-<gitsha>` tag.

## Three things worth reviewing

**Landing order is mandatory.** The path filter includes the workflow's own file, so merging this before the component stack reaches `main` triggers a run against a `main` with no `components/grid-wallet-prod/` directory, and the build fails on a missing working directory. **Land the component stack first, this second.**

**`linux/arm64`** because the tooling-prod cluster is Graviton (m8g). No other workflow in this repo uses an ARM label yet — if the first run sits in `queued`, GitHub-hosted ARM runners aren't enabled for the org and the fallback is `setup-qemu` + `setup-buildx` on `ubuntu-latest` (noted inline). ARC runners are *not* an option here: they have no Docker daemon.

**The sandbox flag is a build arg, not config.** `NEXT_PUBLIC_*` is inlined by Next at build time, so `--build-arg NEXT_PUBLIC_GRID_SANDBOX=true` is written literally in the workflow and has **no default** in the Dockerfile. A future production image would be a separate workflow against a separate ECR repo and cannot inherit the sandbox affordance by accident.

## Idempotency

The ECR repo is `IMMUTABLE`-tagged, so re-pushing an existing tag is a hard failure. The build is guarded by a `describe-images` check that distinguishes *"tag absent"* from *"the call failed"* — a bare `if aws ... >/dev/null 2>&1` is exempt from `set -e` and would treat an IAM error as a missing image, then die on the immutable push. Only `ImageNotFoundException` means "go build"; anything else fails loudly.

No `workflow_dispatch`: the IAM trust policy pins the `sub` claim to `refs/heads/main`, so a dispatch from any other ref could not authenticate.

## Depends on

`tooling-infra` PR (ECR repo + `github-actions-grid-wallet-prod` role) must be applied before this runs, or the first push has nothing to authenticate against.

### DIFF
--- a/.github/workflows/grid-wallet-prod-image.yml
+++ b/.github/workflows/grid-wallet-prod-image.yml
@@ -24,16 +24,22 @@ jobs:
     # runners are not enabled for the org and the fallback is
     # docker/setup-qemu-action + docker/setup-buildx-action on ubuntu-latest.
     runs-on: ubuntu-24.04-arm
+    # Actions are pinned to full commit SHAs, not major tags. This job holds
+    # id-token: write and assumes a role that can push to ECR, so an upstream
+    # tag move would silently change code running with cloud write access.
+    # NOTE: this repo has no Dependabot config, so these pins do not update
+    # themselves — re-resolve them when bumping, or add a dependabot.yml with
+    # a github-actions ecosystem entry so they stay current.
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
 
-      - uses: aws-actions/configure-aws-credentials@v4
+      - uses: aws-actions/configure-aws-credentials@7474bc4690e29a8392af63c5b98e7449536d5c3a # v4.3.1
         with:
           role-to-assume: arn:aws:iam::405785876631:role/github-actions-grid-wallet-prod
           aws-region: us-west-2
 
       - id: ecr
-        uses: aws-actions/amazon-ecr-login@v2
+        uses: aws-actions/amazon-ecr-login@d539f0932e70871a027e9d5a9d8fc38589180a64 # v2.1.6
 
       - name: Build and push
         working-directory: components/grid-wallet-prod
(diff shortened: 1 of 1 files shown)


mariano-macri:
@greptileai review

No new commits since your last review — re-triggering for a fresh pass now that the sibling PRs have changed.

On the mutable action tags you flagged: keeping `@v4`/`@v2` deliberately. All three are first-party (`actions/checkout` from GitHub; `configure-aws-credentials` and `amazon-ecr-login` from AWS), and this repo has no Dependabot config — so SHA pins would have no update path, freeze, and stop receiving upstream security fixes. That's a worse posture than tracking the major tag, not a better one. Pinning is right once there's a mechanism keeping pins current; that's a repo-wide change worth doing deliberately rather than inside a CI-wiring PR.

mariano-macri:
@greptileai review

Rebased along with #746 — commits are now `c6a8a5e` and `6a45f19`, content unchanged (1 file, +81).

mariano-macri:
@greptileai review

Addressed the mutable-action-reference finding in `61bf04b1` — all three actions are now pinned to full commit SHAs with version comments:

```
actions/checkout                      @11d5960 # v4.4.0
aws-actions/configure-aws-credentials @7474bc4 # v4.3.1
aws-actions/amazon-ecr-login          @d539f09 # v2.1.6
```

Reversing my earlier position: this is the first workflow in the repo to hold cloud write credentials (`id-token: write` + ECR push), so it's worth deviating from the repo-wide major-tag convention here.

One caveat noted inline in the workflow: this repo has no `dependabot.yml`, so these pins won't update themselves. A `github-actions` ecosystem entry would keep them current repo-wide — worth doing separately rather than in this PR.

pengying (approved):

PR #756, opened 2026-07-28
ci(grid-wallet-prod): deploy the pushed image via ArgoCD

mariano-macri:
The workflow built and pushed an image, but nothing deployed it. The app has no `syncPolicy.automated` (and the `public-http-app` module doesn't expose one), so every deploy was a manual `argocd sync`. This closes that.

**Depends on lightsparkdev/tooling-infra#138 — land that first.** It grants this repo's main-branch CI the ArgoCD permission used here; without it the deploy job fails with `PermissionDenied`.

## Approach

Follows the pattern golinks / docket / zeroconf / observatory already use against this cluster: GitHub OIDC → Dex token exchange (no long-lived ArgoCD token in repo secrets), set the `image.tag` helm parameter on the Application, sync, wait, verify.

## Details worth a look

- **The `tag` output comes from its own `meta` step.** "Build and push" `exit 0`s early when the immutable tag already exists, so a tag computed *inside* that step would be empty on exactly the re-run path where the deploy still has work to do.
- **`sync` and `wait` are both whole-app.** golinks scopes both to the Deployment; widening the sync (so a chart-only change with no new image still rolls out) while leaving the wait Deployment-scoped would let a failing Service sync through unnoticed.
- **A verify step asserts a live image carries the pushed tag.** `argocd app set --parameter` accepts an unknown key *silently*, so a typo'd parameter name would sync cleanly, leave the old image running, and still pass the health wait — a green build that deployed nothing. This also catches the non-atomic case where `set` landed but the rollout didn't.
- **`if: github.ref == 'refs/heads/main'` on `build`.** The ECR trust policy and ArgoCD RBAC already reject a non-main dispatch, but both live in other repos and could be relaxed independently; this keeps the gate visible next to the trigger. `deploy` inherits it via `needs: build`.
- **No `actions/github-script`.** golinks needs it only to reach `ACTIONS_RUNTIME_TOKEN`, which isn't exposed to `run` steps; `ACTIONS_ID_TOKEN_REQUEST_{URL,TOKEN}` are. One less pinned third-party action in a job that can write to the cluster.
- **No GitHub `environment`, deliberately.** Declaring one rewrites the OIDC `sub` away from `ref:refs/heads/main` and breaks both the ECR role trust policy and the ArgoCD RBAC entry.
- **`timeout-minutes` on both jobs.** No other job in this repo targets an ARC label, so scheduling on `arc-runner-1` is unproven; without a bound, a scale set that never provisions leaves the job queued instead of failing.

## Known tradeoff

Deploys are now automatic on every qualifying merge, which weakens the design spec's accepted-risk mitigation for `Recreate`-induced webhook loss ("don't deploy while someone is demoing") — no human picks the moment now. Accepted knowingly; a GitHub `environment` approval gate was rejected because of the OIDC `sub` constraint above. The escape hatch, if it bites, is gating `deploy` on the commit message.

## Review

Greptile 5/5, no comments. Also went through a 7-review Codex + Gemini panel (correctness / security / tests / spec-compliance / pattern-fit); the fixes above are what came out of it.

### DIFF
--- a/.github/workflows/grid-wallet-prod-image.yml
+++ b/.github/workflows/grid-wallet-prod-image.yml
@@ -6,6 +6,13 @@ on:
     paths:
       - 'components/grid-wallet-prod/**'
       - '.github/workflows/grid-wallet-prod-image.yml'
+  # Manual runs exist so this workflow can be exercised deliberately rather than
+  # only ever executing for the first time against main. Safe for auth: a
+  # dispatch is only offered on a branch, and from main the OIDC sub is still
+  # repo:lightsparkdev/grid-api:ref:refs/heads/main — which is what BOTH the
+  # ECR push role's trust policy and the ArgoCD RBAC entry match on. Dispatching
+  # from any other branch changes the sub and fails to authenticate, by design.
+  workflow_dispatch:
 
 permissions:
   id-token: write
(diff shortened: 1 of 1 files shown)


mariano-macri:
* **#756** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/756?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/756?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

markatto (approved):

PR #784, opened 2026-07-31
feat: add optional ownershipType to external account schema

shreyav:
Adds an optional `ownershipType` field to the external account resource, so external accounts can indicate whether they belong to the customer themselves (`FIRST_PARTY`) or to a third party (`THIRD_PARTY`).

## Changes

- **New enum** `openapi/components/schemas/external_accounts/OwnershipType.yaml` with values `FIRST_PARTY` and `THIRD_PARTY`.
- **`ExternalAccount.yaml`**: adds an optional top-level `ownershipType` property, alongside the other resource-level fields (`customerId`, `currency`, `platformAccountId`, ...).
- **`ExternalAccountCreateRequest.yaml`** and **`PlatformExternalAccountCreateRequest.yaml`**: add the same optional `ownershipType` property so it can be set at creation time for both customer and platform-owned accounts.
- Rebundled `openapi.yaml` and `mintlify/openapi.yaml` via `make build`.

The field lives on the resource envelope rather than inside the per-rail `accountInfo` payloads, matching where the other account-level metadata sits. It is optional, so this is backwards compatible and `info.version` is unchanged.

## Testing

- `make build` — bundles cleanly.
- `npx @redocly/cli lint openapi.yaml` — valid (warnings only, all pre-existing).

Note: the `npx spectral lint` step of `make lint` fails in this environment because the `spectral` binary isn't resolvable, and `make lint-markdown` fails because `package.json` has no `lint:markdown` script. Both are pre-existing and unrelated to this change.

### DIFF
--- a/mintlify/openapi.yaml
+++ b/mintlify/openapi.yaml
@@ -16768,6 +16768,13 @@ components:
         - CNY_ACCOUNT
       description: Type of external account or wallet
       example: AED_ACCOUNT
+    OwnershipType:
+      type: string
+      enum:
+        - FIRST_PARTY
+        - THIRD_PARTY
+      description: Whether the external account belongs to the customer themselves (first party) or to someone else (third party)
+      example: FIRST_PARTY
     BaseExternalAccountInfo:
       type: object
       required:
@@ -16775,6 +16782,8 @@ components:
       properties:
         accountType:
           $ref: '#/components/schemas/ExternalAccountType'
+        ownershipType:
+          $ref: '#/components/schemas/OwnershipType'
     AedBeneficiary:
       title: Individual Beneficiary
       type: object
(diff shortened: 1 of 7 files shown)


jklein24:
Is the main reason to put this top-level just because we don't have a base crypto wallet account type? This is currently only needed for crypto wallets, right? Do you anticipate it might be used or needed for some fiat rails too?

Also, how will people know when this is required or not? Should we add something to docs or the field desription?

shreyav:
> Is the main reason to put this top-level just because we don't have a base crypto wallet account type? This is currently only needed for crypto wallets, right? Do you anticipate it might be used or needed for some fiat rails too?
> 
> Also, how will people know when this is required or not? Should we add something to docs or the field desription?

we do have a base but for some reason I couldn't figure out all the other fields are at this level and only the discriminator is on the base, so i put it here for consistency -- I don't feel strongly about it and can change it if you do. We also need for fiat in some cases, will share context offline

I haven't figure out exactly when its required yet, will update the description when I do

AaryamanBhute (approved):

PR #803, opened 2026-08-05
feat: accept compressed clientPublicKey on passkey challenge + oauth verify

carsonp6:
## Why
Knob-ON passkey/oauth login (`STAMP_LOGIN` / `OAUTH_LOGIN`) registers the client-supplied `clientPublicKey` directly as the session signing key instead of sealing an `encryptedSessionSigningKey` back to it. That session key is stored and matched in **compressed** SEC1 form, so a client that sends the **uncompressed** key fails the first session-stamped call with `401 "Invalid wallet signature."` against real infra. (Sandbox accepts both, so it only bit production; OTP login already sends compressed and works.)

The `AuthCredentialChallengeRequest.clientPublicKey` schema pins `pattern: ^04[0-9a-fA-F]{128}$` + `minLength/maxLength: 130`, so the generated SDKs reject a compressed key outright for the passkey challenge.

## What
- `AuthCredentialChallengeRequest.clientPublicKey`: accept either encoding — `pattern: ^(04[0-9a-fA-F]{128}|0[23][0-9a-fA-F]{64})$`, `minLength: 66`. Description broadened to describe both encodings.
- `OauthCredentialVerifyRequestFields.clientPublicKey`: description broadened to note the compressed encoding is accepted (no pattern was set, so this is doc-only).
- The legacy HPKE-target bodies (`InternalAccountExportRequest`, `AuthSessionRefreshRequest`) are **unchanged** — they still require the uncompressed 65-byte point they seal credentials to.

Backwards-compatible: existing uncompressed clients stay valid.

## Downstream
The webdev backend already accepts both encodings server-side (hand-rolled validator). After this merges, the vendored SDK in `lightsparkdev/webdev` (`grid-api/`) needs a regen (`./update_schema.sh`) so the generated model validators pick up the relaxed constraint. Left the `info.version` (`2025-10-13`) untouched — bump per the usual release process if required.


carsonp6:
* **#803** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/803?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/803?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #806, opened 2026-08-06
feat: add external account ownership verification

shreyav:
## Summary

Adds **ownership verification** for `FIRST_PARTY` self-custody crypto wallet external accounts, as account-scoped challenge/verify actions (per review feedback — this replaces the earlier standalone `/ownership-verifications` resource design).

### Endpoints (on both `/customers/external-accounts/{id}` and `/platform/external-accounts/{id}`)
- `POST …/challenge` — start (or restart) verification with a chosen `method`:
  - `WALLET_SIGNATURE` → returns a `messageToSign`
  - `LIVENESS` → returns a hosted `verificationLink` (+ optional embed `token`, mirroring the KYC link contract)

  Calling `challenge` again abandons any in-flight challenge — that's how you retry after a failure, replace an expired challenge, or switch methods. An `UNVERIFIED` account returns to `PENDING_OWNERSHIP_VERIFICATION`.
- `POST …/verify` — complete a `WALLET_SIGNATURE` challenge synchronously by submitting the signature (`signatureScheme`: `bip137`/`electrum` for Bitcoin). `LIVENESS` completes asynchronously and never calls this endpoint.

Nothing verification-related is collected on the account create request.

### External account statuses
- `PENDING_OWNERSHIP_VERIFICATION` — ownership must be verified before unrestricted use (e.g., under EU Travel Rule or similar regimes). Covers both "not yet started" and "in progress" (liveness underway). Usable below regulatory thresholds.
- `UNVERIFIED` (new) — the most recent verification attempt failed; same below-threshold capabilities; a new challenge returns it to `PENDING_OWNERSHIP_VERIFICATION`.
- Success → `ACTIVE`.

Both statuses only ever appear on crypto wallet accounts on platforms with counterparty requirements, so existing integrations are unaffected.

### Webhooks
- `EXTERNAL_ACCOUNT.STATUS_UPDATED` (new, mirrors `INTERNAL_ACCOUNT.STATUS_UPDATED`) fires on any status transition and carries the full account. Because failed attempts are a status transition (`→ UNVERIFIED`), this single event covers the whole verification lifecycle — including async liveness failures — with no bespoke verification webhooks.

### Error codes
- Renamed `WALLET_VERIFICATION_REQUIRED` to `EXTERNAL_ACCOUNT_VERIFICATION_REQUIRED`; removed `OWNERSHIP_TYPE_REQUIRED`, `THIRD_PARTY_TRANSFER_LIMIT_EXCEEDED`, and `LIGHTNING_TRANSFER_LIMIT_EXCEEDED`.

### Other
- Expanded `ownershipType` docs (generalized beyond EU Travel Rule).
- Stainless: `challenge`/`verify` methods on the customer and platform external-accounts subresources + challenge/request models.

## Testing
`make build` bundles cleanly; `redocly lint` and `spectral lint` match the pre-existing baseline on `main` exactly (no new findings).


openapi/openapi.yaml
```diff
@@ -193,6 +197,12 @@ paths:
     $ref: paths/platform/platform_external_accounts.yaml
   /platform/external-accounts/{externalAccountId}:
     $ref: paths/platform/platform_external_accounts_{externalAccountId}.yaml
+  /ownership-verifications:
```
jklein24: Did you consider scoping this to the external account resource? I know it makes the path long, but it does seem pretty tied to the account... `/customers/external-accounts/{externalAccountId}/verification` or something?

openapi/components/schemas/external_accounts/ExternalAccountStatus.yaml
```diff
@@ -2,6 +2,15 @@ type: string
...
   - ACTIVE
+  - PENDING_OWNERSHIP_VERIFICATION
   - UNDER_REVIEW
   - INACTIVE
-description: Status of an external account
+description: |
+  Status of an external account.
+
+  `PENDING_OWNERSHIP_VERIFICATION` applies to `FIRST_PARTY` self-custody
+  crypto wallet accounts on platforms subject to EU Travel Rule requirements.
```
jklein24: nit: It's not just EU TR now right? If we're talking about some ROW regions needing verification too, it might expand...

openapi/openapi.yaml
shreyav: hmm I tried to follow the KYC verification pattern... open to either way

openapi/components/schemas/external_accounts/ExternalAccountStatus.yaml
shreyav: updated!

pengying (approved):

PR #828, opened 2026-08-13
feat: add beneficiary to crypto wallet external accounts

shreyav:
## Summary

Crypto-wallet external accounts were the only account family without a `beneficiary`. This PR adds one, following the fiat pattern — an `INDIVIDUAL`/`BUSINESS` `oneOf` discriminated by `beneficiaryType`.

The beneficiary identifies who owns the wallet — the counterparty identity needed for Travel Rule data, independent of custody.

This is PR 2 of 3 for VASP counterparty support (Travel Rule):
1. #827 — `/vasps` directory
2. **This PR** — `beneficiary` on crypto-wallet external accounts
3. #829 — `custodyType`/`vaspName` on external accounts

### Field requirements: exactly what is transmitted
The individual variant is a new `WalletIndividualBeneficiary` with only **`fullName` + `countryOfResidence`** (both required) — the exact set transmitted as Travel Rule counterparty info. The generic `IndividualBeneficiary` couldn't be reused because it *requires* `birthDate`/`nationality`, which are never transmitted for wallets and would force platforms to collect a third party's date of birth (OpenAPI composition can't relax `required`). Optional PII fields were deliberately omitted: adding optional fields later is non-breaking, while accepting-but-ignoring PII invites needless collection. The `BUSINESS` variant reuses the existing `BusinessBeneficiary` (`legalName` required).

### Semantics (one deliberate divergence from fiat)
- Fiat accounts require `beneficiary`. For wallets it is **optional for `FIRST_PARTY`** — when omitted, the customer's verified identity is used, so the dominant own-wallet case sends nothing extra.
- **Required for `THIRD_PARTY`** wallets on platforms subject to counterparty requirements (e.g., EU Travel Rule and similar regimes) — enforced at runtime with `400 INVALID_INPUT`, not in the schema, since the requirement is platform-dependent.

### Changes
- New `WalletBeneficiaryFields` fragment (the `beneficiary` property) composed into all seven wallet variants: `BASE_WALLET`, `ETHEREUM_WALLET`, `POLYGON_WALLET`, `PLASMA_WALLET`, `SOLANA_WALLET`, `SPARK_WALLET`, `TRON_WALLET`
- New `WalletBeneficiaryOneOf` — the named individual/business union (matches the `*OneOf` house convention)
- New `WalletIndividualBeneficiary` schema (`fullName` + `countryOfResidence`)
- Stainless model entries for all three

### Out of scope
- `LIGHTNING` external accounts — Travel Rule counterparty identity for Lightning flows in-band (payment-level), not via a stored account beneficiary. Flagging in case reviewers feel otherwise.

## Testing
`make build` bundles cleanly; `redocly lint` and `spectral lint` match the pre-existing baseline on `main` exactly (no new findings).


openapi/components/schemas/common/WalletIndividualBeneficiary.yaml
```diff
@@ -0,0 +1,37 @@
...
+  nationality:
+    type: string
+    description: The nationality of the beneficiary (ISO 3166-1 alpha-2)
+    example: US
+  email:
+    type: string
+    description: The email of the beneficiary
+  phoneNumber:
+    type: string
+    description: The phone number of the beneficiary
```
jklein24: When might these 4 be required?

openapi/components/schemas/external_accounts/WalletBeneficiaryOneOf.yaml
```diff
@@ -0,0 +1,12 @@
...
+oneOf:
+  - title: Individual Beneficiary
+    $ref: ../common/WalletIndividualBeneficiary.yaml
+  - title: Business Beneficiary
+    $ref: ../common/BusinessBeneficiary.yaml
+discriminator:
+  propertyName: beneficiaryType
+  mapping:
+    INDIVIDUAL: ../common/WalletIndividualBeneficiary.yaml
+    BUSINESS: ../common/BusinessBeneficiary.yaml
```
pengying, line 12: does this mean there'll be a business beneficiary as well?

pengying: Er I mean is there a individual beneficiary we can resuse

openapi/components/schemas/external_accounts/WalletBeneficiaryFields.yaml
```diff
@@ -0,0 +1,10 @@
+type: object
+properties:
+  beneficiary:
+    description: >-
+      The person or business that owns this wallet. Optional for
```
jklein24: nit: It might be good to indicate the error type that will be returned if this is required, but not provided. That way, people can handle it correctly.

openapi/components/schemas/common/WalletIndividualBeneficiary.yaml
shreyav: currently not required, but added these so they match the fiat individual beneficiary, happy to drop we can always add later

openapi/components/schemas/external_accounts/WalletBeneficiaryOneOf.yaml
pengying: Huh i guess not I see a bunch of different types of beneficiaries depending on the country but not a generic one. 

Do you think this one will be generic or is it only for wallets?

pengying (approved):

openapi/components/schemas/external_accounts/WalletBeneficiaryOneOf.yaml
shreyav: yeah we went through the exercise of splitting up beneficiaries for individuals based on region reqs but we havent done that for business yet -- so reusing the existing business one and added a new one just for wallets w the minimum requirements

PR #850, opened 2026-08-20
Add 200/PROCESSING (WalletOperationProcessing) response to all 6 still-processing Global Accounts endpoints

carsonp6:
## Summary

Grid's embedded-wallet auth endpoints forward to Turnkey, whose consensus- or approval-gated activities are only optimistically synchronous. This adds a documented `200`/`PROCESSING` response option so a still-in-flight activity doesn't have to surface as a 5xx:

- `POST /auth/credentials` (add credential): new `200` response, body `WalletOperationProcessing` (`status: "PROCESSING"`). Client re-sends the byte-identical stamped retry (same `Request-Id`) until the credential is added.
- `POST /auth/credentials/{id}/challenge` (re-issue challenge): `200` response widened from a bare `AuthCredentialResponseOneOf` to a new `AuthCredentialChallengeResponse` wrapper — `anyOf: [AuthCredentialResponseOneOf, WalletOperationProcessing]` — so a still-processing OTP send is now representable.

`WalletOperationProcessing` is placed in `components/schemas/common/` (not `auth/`) because upcoming PRs reuse the identical shape on other Embedded Wallet endpoints (contact update, session refresh, OTP verify) as part of the Turnkey sync-state-machine work (SP-3579).

`anyOf` rather than `oneOf` on `AuthCredentialChallengeResponse`: `WalletOperationProcessing`'s `status` field isn't a discriminator shared with `AuthCredentialResponseOneOf`'s `type` field, so a discriminated `oneOf` doesn't fit — this repo's spectral rules require a discriminator on every `oneOf` and flag inline (non-`$ref`) response schemas, both satisfied here.

## Test plan

- [x] `make build` — rebundled `openapi.yaml` + `mintlify/openapi.yaml` from `openapi/` sources, committed
- [x] `make lint` — 0 errors (was 2: `no-inline-response-schema`, `oneOf-must-have-discriminator`, both fixed by wrapping in a named `$ref`d `anyOf` schema instead of an inline `oneOf`); remaining warnings/infos are pre-existing and unrelated to this change

This is the first step of a grid-api-first rollout: after this merges, the webdev repo's vendored `grid-api/` Python client gets regenerated (`update_schema.sh`) in a separate PR, bumping `packageVersion`, before any consumer code lands on top.

---

PATCH /customers excised — contact updates move to dedicated contact-changes endpoints (design in progress).


openapi/paths/auth/auth_credentials.yaml
```diff
@@ -90,6 +90,18 @@ post:
...
+        An additional-credential setup leg's underlying wallet-provider
+        activity is still in flight — a `WalletOperationProcessing` body
+        with `status: "PROCESSING"`. The client re-sends the
+        byte-identical stamped retry (same `Request-Id`) until the
+        credential is added; the backend also reconciles the activity
+        to terminal on its own.
+      content:
+        application/json:
+          schema:
+            $ref: ../../components/schemas/common/WalletOperationProcessing.yaml
```
shreyav, line 104: should we also send webhooks here so clients dont have to poll? im not sure which resource / id we would send the webhook on tho since theres no id here yet

carsonp6: Good question. Worth splitting by endpoint: for the other 5 of the 6 touched here (challenge/verify, session revoke/refresh), a credential or session `id` already exists at the point PROCESSING is returned, so a resource-keyed webhook is possible later if we want one. This create-credential endpoint is the one case where there's genuinely no `id` yet — the credential doesn't exist until the activity settles, so any notification here would have to key off the client's `Request-Id` rather than a resource id.

Adding webhooks alongside polling for these auth operations is a bigger scope call than this PR (webhook coverage for this surface is being staged separately), so I don't want to fold it in here. Flagging for Carson to weigh in on whether/when to add it rather than deciding unilaterally.

carsonp6: Decision (Carson): staging webhooks separately — we'll likely want them for these ops eventually (the webhook infra already exists), but not in this PR. Polling via the 200/PROCESSING re-send stays the contract here.

shreyav: discussed offline, we'll make this stateful later

shreyav (approved):

PR #853, opened 2026-08-21
Add contact-changes endpoints for customer email/phone updates

carsonp6:
Adds dedicated endpoints for changing a customer's email or phone number, replacing the contact-update half of `PATCH /customers` that was excised from #850. Contact changes are now the **single door** for contact updates: `email` and `phoneNumber` on `PATCH /customers` are marked deprecated here (still functional during migration; nothing removed).

## Why

The email behind `EMAIL_OTP` and the phone number behind `SMS_OTP` are what the customer logs in with. Changing either one is a login-security operation, not a profile edit: it needs the customer's own signature, and it has to re-key every tied OTP credential across every tied wallet as one operation.

`PATCH /customers` carried that as a `202` challenge inside a single request/response pair, which has three problems:

- **The challenge was unrecoverable.** Lose the `202` body — page reload, dropped connection, a backend that didn't persist it — and there is no way to get `payloadToSign` back. The customer starts over.
- **A pending change was invisible.** Nothing to read, nothing to cancel, no way to answer "is a contact change in flight for this customer?" or to show the customer what they already requested.
- **It was welded to a profile edit.** One endpoint that usually completes synchronously and occasionally turns into a two-step signed ceremony, discriminated by whether the body happened to contain `email` or `phoneNumber`.

Modelling the change as a resource fixes all three: the challenge lives on something the platform can `GET`, cancel, and list.

Making it the *only* door fixes a fourth: whether a signature is required is Grid's business, not the integrator's. A contact backing an OTP credential needs one; a contact that isn't backing a credential is just a field. Callers hit one endpoint and branch on the returned `status` instead of tracking which of their customers have which credentials.

## The lifecycle

```mermaid
stateDiagram-v2
  [*] --> AWAITING_SIGNATURE: POST /contact-changes → 202 (tied OTP credential exists)
  [*] --> APPLIED: POST /contact-changes → 201 (no tied credential — applied on create)
  [*] --> FAILED: POST /contact-changes → 201 (no tied credential — inline apply failed)
  AWAITING_SIGNATURE --> PROCESSING: POST .../submit (stamp accepted, provider in flight)
  AWAITING_SIGNATURE --> APPLIED: POST .../submit (settles inline)
  AWAITING_SIGNATURE --> CANCELLED: DELETE
  AWAITING_SIGNATURE --> EXPIRED: expiresAt passes
  PROCESSING --> APPLIED: provider settles
  PROCESSING --> FAILED: provider fails / tied credential sync fails
  APPLIED --> [*]
  FAILED --> [*]
  EXPIRED --> [*]
  CANCELLED --> [*]
```

In prose: a create arrives in one of three states. With a tied OTP credential of that type it arrives `202 AWAITING_SIGNATURE` holding `payloadToSign` and `expiresAt`; the client stamps that payload with the session key of any verified credential on one of the customer's tied wallets and submits it, and the submit either settles inline to `APPLIED` or comes back `PROCESSING`, where re-sending the identical stamp converges on the terminal answer. With no tied credential of that type there is nothing to re-key and no signature to collect, so Grid applies the change on create and it arrives `201` and terminal with neither field — `APPLIED` on success, or `FAILED` with a `failureReason` if the inline apply didn't work. Either way it's recorded, and either way there's nothing further to call.

Grid reconciles a `PROCESSING` change on its own regardless, so a client that stops retrying can read the outcome from `GET`. An unsubmitted change can be cancelled deliberately, and lapses to `EXPIRED` on its own otherwise. Once submitted there is nothing to cancel: we don't pretend to un-sign a signature.

`APPLIED` means the customer contact field *and* every tied matching OTP credential were updated. If any tied credential can't be updated, the contact field is left alone and the change ends `FAILED` — nothing lands half-applied.

## Endpoints

| Endpoint | Purpose |
|---|---|
| `POST /customers/{customerId}/contact-changes` | Create a change. `202 AWAITING_SIGNATURE` + `payloadToSign` when a tied OTP credential exists; `201 APPLIED` or `201 FAILED` when none does, since Grid applies it inline. One active change per contact type — a second create while one is active returns `409 CONTACT_CHANGE_PENDING` with `details.contactChangeId` naming the live one, and no resource. |
| `POST /customers/{customerId}/contact-changes/{changeId}/submit` | Carry the stamp in `Grid-Wallet-Signature`. `200` with the `APPLIED` change, or `200 WalletOperationProcessing` while in flight. Idempotent. |
| `GET /customers/{customerId}/contact-changes` | List, newest first, paginated, filterable by `status`. |
| `GET /customers/{customerId}/contact-changes/{changeId}` | Read one — including re-reading `payloadToSign` after a lost create response. |
| `DELETE /customers/{customerId}/contact-changes/{changeId}` | Cancel while `AWAITING_SIGNATURE`; `409` otherwise. |

Status enum: `AWAITING_SIGNATURE`, `PROCESSING`, `APPLIED`, `FAILED`, `EXPIRED`, `CANCELLED`.

## Based on #850

This branch is stacked on `sp3579-wallet-operation-processing`, because `submit` returns #850's `WalletOperationProcessing` on the in-flight path. Review #850 first; this PR's diff against it is only the contact-change surface plus the PATCH deprecation notes.

## Lint / build

`make lint` passes: **0 errors**, 150 warnings (baseline on the parent branch was 148). The two added warnings are both known rule quirks, neither a real finding:

- `delete-returns-204` on the cancel endpoint — deliberate, see open question 9. `DELETE /customers/{customerId}` already returns `200` + the resource.
- `pagination-envelope-has-data` on `ContactChange` — the rule fires on any `GET` `200` schema; `ContactChange` is the read-one response, not a list. Same false positive already exists for `StablecoinOperation` and `StablecoinProviderAccount`.

`make build` regenerated `openapi.yaml` and `mintlify/openapi.yaml`; both are committed in sync.

Note that **Lint, OpenAPI Build, and breaking-changes do not run on this PR** — those workflows are scoped to `pull_request: branches: [main]`, so a stacked PR skips them. The gate that does run is `preview` (the Stainless SDK build off `openapi.yaml` + `.stainless/stainless.yml`), which passes. The other three fire for the first time when this retargets main after #850 merges.

## Decided

**Status codes: `202` when a signature is required, `201` when it isn't.** ✅ Ruled: the house convention holds — `202` whenever Grid needs the client's signature before proceeding. So an `AWAITING_SIGNATURE` arrival is `202` with the `ContactChange` as the body (identical fields, only the code changed), and terminal arrivals stay `201`. The resource-body part of the design is unchanged.

The two codes answer different questions, which the endpoint now states outright: `202` means accepted and waiting on the customer's signature; `201` means created and settled with nothing further required of the caller — and "settled" includes a recorded `FAILED` attempt, so a `201` still means the change was recorded rather than that it worked; `4xx` stays refused-and-unrecorded.

A side benefit: create now matches every other Embedded Wallet endpoint that needs a signature first, so there's no new convention for integrators to learn. It also retires the old "unlike the `202` on `PATCH /customers`" framing, which no longer distinguished anything once both returned `202`. The difference that survives is the one that always mattered: the body is a change resource whose `payloadToSign` can be re-read after a lost response, not a bare challenge that exists only in that one response.

**6. A create for a customer with no tied OTP credential no longer `400`s — it returns `201` already `APPLIED`.** ✅ Decided: contact-changes is the only door. Grid applies the profile change and downstream sync inline on create, with the same semantics as today's PATCH path, and the change arrives terminal with no `payloadToSign` and no `expiresAt`. Every arrival state is documented on the create endpoint, and callers are told to branch on `status` rather than on what they believe about the customer's credentials. The dead `400` clause is gone; every other error code is unchanged. `PATCH /customers` keeps `email` and `phoneNumber` working, now marked deprecated in the endpoint description, the `CustomerUpdateRequest` schema description, and both field descriptions.

One contract hole this opened, closed in a follow-up commit: a change can now reach `APPLIED` having never had a `payloadToSign`, so submitting it was ambiguous — the `200` clause said an `APPLIED` change returns its own body, while the `401` clause said a stamp not matching `payloadToSign` is unauthorized, and both read as applying. Resolved toward the blind retry being safe: an applied change has nothing to verify against and nothing to mutate, so the stamp isn't checked and the call returns the applied change either way. Flagging in case you'd rather it `409`'d to make "you never needed to sign this" explicit.

**16. A failed instant-apply is recorded, not discarded.** ✅ Decided: a create whose inline apply fails returns `201` with the change persisted as `FAILED` plus a `failureReason`, so every attempt is visible in the list whether or not it worked. Create therefore has three arrival states — `AWAITING_SIGNATURE` with a tied OTP credential, `APPLIED` or `FAILED` without one — documented on the endpoint description, both success-response descriptions, `ContactChange`, `ContactChangeStatus`, and the auth guide, with a `failedOnCreate` example.

The status code follows repo precedent rather than instinct alone, since you asked me to look: `POST /agents/{agentId}/actions/{actionId}/approve` and `POST /agents/me/quotes/{quoteId}/execute` both return `2xx` carrying an `AgentAction` that may be `FAILED`, and reserve `4xx` for a request that can't be accepted. `AgentActionStatus` draws the same line in its own vocabulary — `REJECTED` for a request turned down, `FAILED` for one that ran and didn't work. So the repo already says "2xx + resource for a failed attempt, `4xx` for a refused request", which is the boundary you described. `201` rather than their `200` because this call creates the resource.

That boundary is now stated at the `409` itself: a rejected request leaves no resource because nothing was attempted, so a duplicate value or an already-pending change of the same type stays a plain `409` with no record. The guide puts it as — a `201` means the change was **recorded**, not that it worked.

**Cross-flow race, fixed here with the mirror agreed elsewhere.** Adding an `EMAIL_OTP` / `SMS_OTP` credential and changing the contact behind one both rewrite the same underlying user attribute — visible directly in this spec, where `POST /auth/credentials` and a contact change publish byte-identical `ACTIVITY_TYPE_UPDATE_USER_EMAIL` payloads. Two submitted at once means whichever settles last silently wins. Create and submit now both `409` with a new `AUTH_CREDENTIAL_OPERATION_IN_FLIGHT` code; on submit the change stays `AWAITING_SIGNATURE`, so a transient overlap never costs the customer their signature.

Scoped deliberately to **submitted** operations: an issued-but-unsigned credential challenge hasn't forwarded anything to the provider, so there's nothing to race, and blocking on it would let two abandoned challenges — one per side — lock a customer out of both flows until they expired, with neither side able to tell "working" from "abandoned". Cross-checked with the pending-credentials design, which holds the mirror guard on the same terms: both sides block only on submitted operations, and neither terminally fails the blocked side. The guide also documents the reverse order, where an applied contact change invalidates an outstanding credential-add challenge because the stamped payload names the old contact.

Related gap for whoever owns the auth endpoints, not fixed here: that staleness rejection is real shipped behavior, and `POST /auth/credentials`' `401` description doesn't mention it — an integrator hitting it has no way to know why. Worth folding into the credential-side change rather than expanding this PR.

## Open questions for review

17. **Should the deprecation be the machine-readable `deprecated: true` flag, not just prose?** You said note, so I wrote notes. Setting `deprecated: true` on the two fields would propagate into every generated SDK as a deprecation warning — correct eventually, but loud while those fields are still the functional path for live integrators. My read: prose now, flip the flag when the contact-changes path has real adoption. Confirm the timing.

Carried over:

1. **Tag / SDK grouping.** All five operations are tagged `Customers` (SDK namespace `customers.contactChanges`), not `Embedded Wallet Auth`. The single-door decision strengthens this: `PATCH /customers` now carries a deprecation note pointing at a sibling in the same tag group, and the endpoint serves customers with no wallet credentials at all, so filing it under an Embedded Wallet tag would be actively wrong for that half of its traffic. Flagging only because the mechanism is still Embedded Wallet vocabulary.
2. **`payloadToSign`, not `payload_to_sign`.** Your sketch used snake_case; spectral enforces camelCase, and this matches the existing `SignedRequestChallenge.payloadToSign`. Same for `expiresAt` / `createdAt`. Flagging only in case you meant a different field.
3. **`CANCELLED`, not `CANCELED`.** Your sketch used one L. The repo's only precedent (`UmaInvitation` status, `INVITATION_CANCELLED`) uses two, so I matched the repo.
4. **`{type, value}` vs `{email}` / `{phoneNumber}`.** Kept your `{type: EMAIL|PHONE, value}` — uniform across list/read, explicit enum, one code path. The alternative mirrors the `Customer` resource's own field names, which integrators may expect. Related: `PHONE` as the enum value, when the customer field is `phoneNumber` and the credential type is `SMS_OTP`.
5. **Submit takes no request body and no `Request-Id`** — just `Grid-Wallet-Signature` over the change's `payloadToSign`. The auth endpoints need `Request-Id` because their challenge exists only in a prior response; here `changeId` in the path is the correlation and the change already holds the fields being changed. Confirm you want the header rather than a `{signature}` body field, and that dropping `Request-Id` is right.
7. **Failure is expressed two ways on the signed path.** A failure Grid sees during submit is `424` (mirroring PATCH's existing `424`) *and* moves the change to `FAILED`. A failure discovered after a `PROCESSING` response has no status code to land on, so it only surfaces as `FAILED` + `failureReason` on the resource. The alternative — submit always returns `200` with the change — gives one place to look but makes `200` mean "your change failed", which fights every SDK's error handling. See also question 16 for the third case, the inline apply on create.
8. ~~**`failureReason` is free text.**~~ ✅ **Resolved — it's a code, and my prose version was the anomaly.** Surfaced by the pending-credentials cross-check, then confirmed against the repo: `StablecoinOperation.failureReason` is "Stable internal failure code" with `PROVIDER_TRANSFER_FAILED`, `OutgoingTransaction.failureReason` `$ref`s a dedicated failure-reason enum, and `mintlify/snippets/error-handling.mdx` documents `failureReason` as the machine token integrators branch on, with lists of `UPPER_SNAKE` values. I had given the same field name a human sentence. Fixed by keeping the name and making it what the name already promises. I did **not** take the proposed `failureCode` + `failureMessage` split — that would be a third convention for one concept, and no existing schema in this repo pairs a code with a prose message (#802's `OperationError` is `{code}` alone). One thing to confirm: I specified the code as **Grid's own vocabulary, not a pass-through of the provider's**, partly because #802's example value is `DeleteApiKeysFailed`, which is PascalCase and reads like provider vocabulary — worth a look there given we don't put provider names or their terminology in the public surface.
9. **Cancel returns `200` + the `CANCELLED` change, not `204`.** Saves a follow-up `GET` to render terminal state; costs a spectral warning. `DELETE /customers/{customerId}` sets the `200` precedent.
10. **List is fully paginated** (`limit`, `cursor`, `hasMore`, `nextCursor`, `totalCount`) plus a `status` filter, unlike `GET /auth/credentials` which is deliberately unpaginated. Contact changes accumulate over a customer's lifetime, so pagination. No `type` filter — say the word if you want one.
11. **Create is not idempotent, and the no-credential path makes that more visible.** A second create with an *identical* `{type, value}` while one is active still `409`s. But a change that arrives `APPLIED` is terminal on arrival and so never blocks the next one — meaning repeated identical creates on the no-credential path each succeed and each leave their own `APPLIED` record. Options: leave it (every call is a real request, honestly recorded), dedupe when `value` already equals the current contact, or return the existing change instead of `409` on the signed path.
12. **Fields added beyond your sketch:** `customerId` and `updatedAt` (matching `AuthMethod`). `payloadToSign` and `expiresAt` appear only while `AWAITING_SIGNATURE` — a consumed payload isn't echoed back, and a change that arrives `APPLIED` never had one.
13. **Five new error codes.** `CONTACT_CHANGE_PENDING` and `CONTACT_CHANGE_NOT_AWAITING_SIGNATURE` (one code for both "can't sign this" and "can't cancel this", naming the required state like `TRANSACTION_NOT_PENDING_PLATFORM_APPROVAL` does), plus the `SMS_OTP` counterparts of codes that only existed for `EMAIL_OTP`: `SMS_OTP_PHONE_ALREADY_EXISTS`, `SMS_OTP_CREDENTIAL_SET_CHANGED`, `SMS_OTP_CREDENTIAL_SYNC_FAILED`. `PATCH /customers` already described those phone cases in prose without ever defining codes for them.
14. **`.stainless/stainless.yml` is not updated,** so these endpoints get no SDK methods yet. `auth/credentials` is registered there; `verify-email`, `verify-phone`, and the SCA trust endpoints are not — spec endpoints that never reached the SDKs. I left it out as outside "hand-edit `openapi/` sources", but this is now the only door for contact updates, so it shouldn't repeat that. Proposed mapping, for this PR or a follow-up: `customers.contact_changes` with `create` / `list` / `retrieve` / `cancel` / `submit`.
15. **No webhook.** A change that settles asynchronously — or fails after a `PROCESSING` response — is discoverable only by polling `GET`. Should there be a `contact_change.applied` / `contact_change.failed` event?


carsonp6:
Closing: we're going with a different API shape for verified contact changes. It reuses POST /auth/credentials (pending credential + /verify) and the existing two-leg PATCH /customers, with no new contact-changes resource. The spec changes will come in a fresh PR.

PR #856, opened 2026-08-23
docs: make /quotes the single transfer path, deprecate /transfer-in and /transfer-out

shreyav:
Same-currency transfers now route through the quote endpoint on the backend. This makes `POST /quotes` the single documented path for moving money, deprecates the platform transfer endpoints, and removes the unbuilt agent ones.

> [!NOTE]
> This PR will pick up the `breaking-change` label — see [Breaking changes](#breaking-changes) below. That is expected and comes from removing the agent transfer endpoints.

## Why

In `sparkcore`, both platform transfer handlers now build a `QuoteRequest` and delegate to `gen_run_create_quote`:

- `sparkcore/grid/api_handlers/transfer_out.py:317` (`_gen_transfer_out_via_quotes`)
- `sparkcore/grid/api_handlers/transfer_in.py:327` (`_gen_transfer_in_via_quotes`)

Both map `amount` onto `lockedCurrencySide: SENDING` / `lockedCurrencyAmount` and set `immediatelyExecute: true`. That mapping is what the migration notes here document.

Two things this deliberately does **not** claim about the platform endpoints:

- The delegation is gated per-platform (`GRID_TRANSFER_IN_VIA_QUOTES` / `GRID_TRANSFER_OUT_VIA_QUOTES`), and they still return a `Transaction` with unchanged request/response shapes. Those are **deprecated, not removed**.
- No sunset date is stated, since none has been set.

## API reference

- `deprecated: true` on `POST /transfer-in` and `POST /transfer-out`, each with a field-by-field migration to `POST /quotes`. This renders a badge in Mintlify and propagates to the generated SDKs via Stainless — no `.stainless/stainless.yml` change needed for these, and removing the resources there would break the SDKs.
- **Inverted the guidance on `POST /quotes`**, which previously read *"If you are transferring funds in the same currency, use the `/transfer-in` or `/transfer-out` endpoints instead."*
- Added a `sameCurrencyAccountToAccount` request example, and updated the Same-Currency / Cross-Currency tag descriptions.

## Removed: agent transfer APIs

`POST /agents/me/transfer-in` and `POST /agents/me/transfer-out` are **removed outright** rather than deprecated. There is no handler for either in sparkcore, so nothing can be using them, and agent transfers will go through quotes when they are built.

Removing the endpoints orphaned the rest of the transfer-shaped agent model, so that goes too:

| Removed | Why it was orphaned |
|---|---|
| `AgentActionType.TRANSFER_OUT` / `TRANSFER_IN` | No endpoint could produce an action of either type |
| `AgentTransferDetails` + `AgentAction.transferDetails` | Only ever populated for those action types |
| `AgentPermission.CREATE_TRANSFERS` | Gated only these two endpoints; `CREATE_QUOTES` / `EXECUTE_QUOTES` cover the quote path |

`.stainless/stainless.yml` and the agent guides (`policies-and-permissions.mdx`, `approvals-and-audit.mdx`) follow.

Worth a reviewer's eye: removing the enum values and the permission goes a step beyond removing the two endpoints. It is the coherent end state if agent transfers become quotes, but it is the most opinionated part of this PR and the easiest piece to scale back.

## Guides

Every transfer-in/transfer-out example is now the equivalent `POST /quotes` call with `immediatelyExecute: true`. The non-obvious part: a quote response is not a transaction, so response payloads and the "track status" steps read `transactionId` off the quote.

**`send-payment.mdx`: 511 lines to 369.** It had two parallel ~140-line walkthroughs, same-currency and cross-currency, that both called `POST /quotes` — a split dating from when same-currency was a genuinely different API. They are now one **Send a payment** flow whose real fork is one-step (`immediatelyExecute`) versus two-step (review the rate, then execute). That is the choice that actually exists, and it is not the same question as whether the currencies differ.

That merge also resolved **two `Transaction statuses` tables that described the same statuses differently** — now one reconciled table. The two-card Overview grid framing "two payment methods" is gone.

**One deprecation notice, not four.** A first pass repeated the same "use `/quotes` instead" note on four pages, so a reader going through core concepts and then a payment guide hit it three times. It now appears once, on the quote system page, where the "when do I need a quote?" question is answered. The API reference badges and the changelog carry the announcement for anyone arriving from the old endpoints.

The transaction lifecycle page's **Same-Currency Transfers** section shrank from ~50 lines to a paragraph, for the same reason: its two request blocks re-showed the lifecycle walked through immediately above them.

## Validation

| Check | Result |
|---|---|
| `make lint` (`lint.yml`) | exit 0 — 0 errors; 663 problems, unchanged from baseline (all pre-existing) |
| `make build` + bundle sync (`openapi-build.yml`) | `openapi.yaml` / `mintlify/openapi.yaml` in sync, rebuild is deterministic |
| oasdiff 1.16.0 (`openapi-breaking-changes.yml`) | 6 errors / 7 warnings — see below |
| Doc anchors | all verified, including relinking the two inbound `#same-currency-transfers` links to `#send-a-payment` |

`mint broken-links` requires a TTY so it could not be run here; it is not a CI gate.

### Breaking changes

oasdiff reports 6 errors, all from the agent removals:

- `POST /agents/me/transfer-in` and `POST /agents/me/transfer-out` — path removed without deprecation
- `CREATE_TRANSFERS` enum value removed from `POST /agents` and `PATCH /agents/{agentId}/policy`
- `TRANSFER_IN` / `TRANSFER_OUT` enum values removed from the `agent-action` webhook

Plus 7 warnings for `transferDetails` disappearing from agent action responses.

The workflow does not fail the job on findings — it posts a sticky comment, adds the `breaking-change` label, and notes that an API reviewer must approve. Since none of these endpoints or fields have a backend implementation, the breakage is theoretical, but the label is correct and the gate should be honored.

## Follow-up: grid-visualizer code generator

`components/grid-visualizer/src/lib/code-generator.ts` still generates `/transfer-out` (L190) and `/transfer-in` (L215) sample code — `canUseTransferOut` / `canUseTransferIn` short-circuit ahead of the quote path at L107-111.

This PR leaves that code alone and only marks the two endpoints **Deprecated** in `components/grid-visualizer/CLAUDE.md`, so the doc stays accurate to what the tool emits today. Routing those branches through the quote path is a behavior change to the visualizer with its own testing surface, and belongs in a separate PR.


openapi/components/schemas/agents/AgentTransferDetails.yaml
pengying, line 1: @k15z heads up we're deprecating the transfer api in favor of unifying everything into quotes

mintlify/platform-overview/core-concepts/transaction-lifecycle.mdx
```diff
@@ -74,46 +74,14 @@ Most transactions on Grid are completed in seconds.
 
 ## Same-Currency Transfers
```
pengying: do we need to break out same currency transfers or can we delete the section?

mintlify/snippets/sending/same-currency.mdx
```diff
@@ -1,6 +1,6 @@
 ## Same-Currency Transfers
 
-Use the `/transfer-out` endpoint when sending funds in the same currency (no exchange rate needed). This is the simplest and fastest option for domestic transfers.
+Use the `/quotes` endpoint when sending funds in the same currency (no exchange rate needed). Quotes cover same-currency and cross-currency transfers alike, so one integration handles both.
```
pengying: i don't know if we still need the distinction for same currency now

pengying (approved):

PR #872, opened 2026-08-27
docs: add a required fields by corridor page

JasonCWang:
Adds a page under Payouts & B2B > Managing accounts that lists, for every
currency and payment rail, the details the sender and the recipient's account
must supply. Readers filter by sender type, recipient type, and currency.

The table itself is generated upstream from the payment integrations and
arrives through the existing schema sync, so it stays current as corridors
change. This PR carries the first copy so the page renders.

Hand-written pieces:
- payouts-and-b2b/depositing-funds/required-fields.mdx, the page that includes
  the snippet
- corridor-fields.js, the filter behaviour
- style.css, the filter bar and empty state, using the existing ls tokens
- docs.json, one navigation entry


JasonCWang:
@greptile review

AaryamanBhute (approved):

PR #880, opened 2026-08-31
chore: sync account schemas, including the create side

JasonCWang:
Regenerated from the schema generator, which now emits the create variant alongside the response variant for every currency.

- ILS is complete: account, beneficiary, business beneficiary, payment account, and both external account variants, plus its entries in the two unions and the account type enum. Creating an ILS external account was not expressible before this.
- Corridors that support no business payee — BDT, COP, EGP, GHS, GTQ, HTG, JMD, PKR — offer only an individual beneficiary, on both the create and the response schema. They previously accepted a business beneficiary on create and returned one the response schema did not allow.
- Every other currency now references its own business beneficiary schema rather than the shared one, on both variants.

CAD, SLV and SWIFT are unchanged: they have no generated counterpart and stay hand-maintained.

Supersedes #876.

### DIFF
--- a/mintlify/openapi.yaml
+++ b/mintlify/openapi.yaml
@@ -14761,9 +14761,9 @@ components:
         bankName:
           type: string
           description: The name of the financial institution holding the account. Optional on every rail, and recommended for wires, where it identifies the beneficiary's institution on the payment message.
-          example: Chase Bank
           minLength: 1
           maxLength: 140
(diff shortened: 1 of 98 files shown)


JasonCWang:
@greptile review

JasonCWang:
@greptile review

whoisglover (approved):

PR #891, opened 2026-09-02
feat: add rule-based internal accounts and the sweep failure webhook

bsiaotickchong:
## Summary

Adds the API surface for **rule-based internal accounts** — an additional account number for an existing customer with a routing rule attached, so incoming payments can be attributed to a specific payer and forwarded automatically.

Every schema change the feature needs is bundled here, in one reviewable PR, rather than arriving in pieces.

## What's added

**`RULE_BASED` on `InternalAccountType`**, plus the same value as a `type` filter on `GET /customers/internal-accounts`.

**`POST /internal-accounts`** — the create verb lives on the bare collection, not under `/customers/`: the body carries an optional `customerId`, so a customer-scoped path would describe a request that may name no customer. The listing paths keep their `GET`s, since a list has to pick a population and a create does not. The body takes `type` and `currency`, an optional `label`, and a `sweepRule` describing where funds are forwarded: a `destination` (account id plus an optional payment rail) and optional `purposeOfPayment`, `description`, and `remittanceInformation`. `Idempotency-Key` is required, matching the other endpoints that mint something irreversible.

Only `RULE_BASED` is creatable. The other account types are provisioned automatically when a customer is created or approved, so the endpoint rejects them with a specific message rather than a generic error.

**`SWEEP.FAILED` webhook** — fired whenever a settled payment does not reach the rule's destination, including when the balance is below the corridor minimum and is returned to the payer instead. The payload carries both transaction ids, a `reason`, and an `outcome`, so an integrator can distinguish "this payment failed" from "and therefore this amount went somewhere else."

Delivery is at-least-once and a redelivery carries a new event id, so the payload documents deduplicating on `incomingTransactionId`.

**`label` and `sweepRule` on the `InternalAccount` response.** A rule-based account can now show what it is: the label recorded at creation, and the rule itself — destination, the derived `minimumAmount` / `maximumAmount` band, purpose, remittance and any fee override. Both are output-only, and the band is derived from the corridor at read time rather than stored, so what a platform reads is what the forward will actually enforce. There is deliberately no `sweepRule.id`: the rule has no lifecycle apart from its account, and publishing an id would invite a resource that does not exist.

## Two decisions worth a second opinion

**`DESTINATION_UNAVAILABLE` is not included.** It appeared in the original design, but nothing in the implementation can produce it — rail validation raises a single condition that `NO_ELIGIBLE_RAIL` already covers. Publishing a value that never arrives costs a permanently un-removable enum member (adding one is non-breaking; removing one is not) and generates a dead case in every SDK. Adding it later, if a rail ever produces it, is free. Happy to reserve it if you'd rather.

**`ABOVE_MAXIMUM` is included** and wasn't in the original design. A balance over the corridor ceiling would otherwise be submitted whole, rejected, and stranded; it now takes the same return path as any other non-success outcome, and this is how the platform is told.

## What's deliberately not here

`GET /customers/internal-accounts/{id}` and `DELETE` were in the original design but are not implemented. Speccing them now would generate SDK methods that 405, so they're left for whenever the endpoints land.

## Verification

`make lint` passes with 0 errors, and zero warnings or informational findings on any schema added here. All 1,971 `$ref`s resolve, none dangling.

Everything is edited under `openapi/` and the bundles regenerated with `make build` — `openapi.yaml` and `mintlify/openapi.yaml` are output, so a source-only change would be reverted by the next build and a bundle-only change would be reverted just as silently. The route move adds a source path file (`openapi/paths/internal_accounts.yaml`) split out of the customers path, and the response fields add two source schemas (`SweepRule.yaml`, `SweepRuleDestination.yaml`).

Original PR: https://github.com/lightsparkdev/grid-api/pull/835


openapi/components/schemas/customers/InternalAccountCreateRequest.yaml
```diff
@@ -0,0 +1,46 @@
...
+type: object
+description: >-
+  Request body for `POST /internal-accounts`. Only `RULE_BASED`
+  accounts can be created directly, and they always carry a `sweepRule`.
+additionalProperties: false
+required:
+  - customerId
+  - type
+  - currency
+  - sweepRule
```
shreyav: can we make this optional so the endpoint can stay generic? i know we can't actually make other accounts rn, but helps us in the future

openapi/components/schemas/customers/SweepRule.yaml
```diff
@@ -0,0 +1,52 @@
...
+      rather than a moving estimate — it changes only when the platform's or
+      Grid's configuration changes, so there is nothing to re-poll.
+  maximumAmount:
+    allOf:
+      - $ref: ../common/CurrencyAmount.yaml
+    description: >-
+      The largest balance the corridor to the destination can carry. Null means
+      no ceiling applies. A settled balance above it is not swept.
+  purposeOfPayment:
+    allOf:
```
shreyav, line 32: curious why we need all of these `allOf`

openapi/components/schemas/customers/SweepRule.yaml
```diff
@@ -0,0 +1,52 @@
...
+      - $ref: SweepRuleDestination.yaml
+    description: Where funds that settle into this account are swept.
+  minimumAmount:
+    allOf:
+      - $ref: ../common/CurrencyAmount.yaml
+    description: >-
+      The smallest balance the corridor to the destination can carry,
+      denominated in this account's own currency. A settled balance below it is
+      not swept. Zero means no floor applies, which is the case for a
```
shreyav: is this denoted in sending currency? would we ever have a minimum for same-currency internal dest because of deposit rail fees?

openapi/components/schemas/webhooks/SweepFailure.yaml
```diff
@@ -0,0 +1,50 @@
+title: Sweep Failure
```
shreyav: hmm im a bit confused why we need this and can't just send a transaction webhook since theyre getting those anyways

bsiaotickchong: ok i updated the PR to no longer send a SweepFailure webhook and instead adds sweepFailureReason and sweepCanonicalAccountId to both incoming and outgoing transactions. Also made the receivedAmount optional for IncomingTransaction based on the discussion we had about creating the operation upon deposit before the quote is created.

openapi/components/schemas/customers/InternalAccountCreateRequest.yaml
shreyav: oh good point to make customer id optional but should we also make sweepRule optional

openapi/components/schemas/customers/SweepRule.yaml
shreyav: can you make the description more concise and obvious that minimumAmount and maximumAmount are denominated in the rule based account currency and not the destination currency

openapi/components/schemas/transactions/IncomingTransaction.yaml
```diff
@@ -29,3 +31,19 @@ allOf:
       failureReason:
         $ref: ./IncomingTransactionFailureReason.yaml
         description: If the transaction failed, this field provides the reason for failure.
+      sweepFailureReason:
```
shreyav: how is this expected to be used alongside failure reason? as an integrator it could be confusing to have to read two fields for rule based accounts

openapi/components/schemas/transactions/IncomingTransaction.yaml
```diff
@@ -14,7 +13,10 @@ allOf:
         $ref: ./TransactionSourceOneOf.yaml
       receivedAmount:
         $ref: ../common/CurrencyAmount.yaml
-        description: Amount received in the recipient's currency
+        description: >-
+          Amount received in the recipient's currency. Absent until the amount
```
shreyav: nit maybe something like this:
Amount received in the recipient's currency. This is only absent for rule-based account sweeps if the sweep couldn't be quoted. It's always present otherwise.

openapi/components/schemas/transactions/IncomingTransaction.yaml
bsiaotickchong: ^ having AI remove this

openapi/components/schemas/transactions/IncomingTransaction.yaml
```diff
@@ -29,3 +31,19 @@ allOf:
...
         description: If the transaction failed, this field provides the reason for failure.
+      sweepFailureReason:
+        allOf:
+          - $ref: ./SweepFailureReason.yaml
+        description: >-
+          Set when this payment settled into a rule-based account but was not
+          swept to the destination of the account's rule. The deposit itself
+          succeeded, so the transaction still completes; this says why the
+          destination was not paid. Omitted for any other account.
+      sweepCanonicalAccountId:
```
shreyav: could we revert this for now? I want to add a more generic "refund destination" soon which i think we could use for this as well

openapi/components/schemas/transactions/IncomingTransaction.yaml
bsiaotickchong: Use shreya's version instead

openapi/paths/customers/customers_internal_accounts.yaml
```diff
@@ -28,8 +28,9 @@ get:
       in: query
       description: >-
         Filter by internal account type. Use `EMBEDDED_WALLET` to find the
-        self-custodial wallet provisioned for a customer, or `INTERNAL_FIAT` /
-        `INTERNAL_CRYPTO` for the platform-managed holding accounts.
+        self-custodial wallet provisioned for a customer, `INTERNAL_FIAT` /
```
shreyav, line 31: do we need same change on the platform path

openapi/components/schemas/transactions/OutgoingTransactionFailureReason.yaml
```diff
@@ -43,3 +48,13 @@ description: |
   | `COMPLIANCE_REJECTED` | The payout partner rejected the recipient on compliance grounds — sanctions, watchlist, or KYC/AML screening |
   | `LIGHTNING_PAYMENT_FAILED` | Deprecated — superseded by `QUOTE_EXECUTION_FAILED`. The Lightning settlement leg failed |
   | `COUNTERPARTY_POST_TX_FAILED` | Deprecated — coarse fallback retained for historical transactions when no granular payout reason is available |
+  | `SWEEP_BELOW_MINIMUM` | The payment settled into a rule-based account but was below the smallest amount the corridor to the rule's destination can carry |
+  | `SWEEP_ABOVE_MAXIMUM` | The payment settled into a rule-based account but was above the largest amount the corridor to the rule's destination can carry |
+  | `SWEEP_NO_ELIGIBLE_RAIL` | No payment rail available to the rule's destination could carry the sweep |
+  | `SWEEP_QUOTE_FAILED` | Grid could not price the sweep to the rule's destination |
+  | `SWEEP_EXECUTION_FAILED` | The sweep was priced but the resulting payment did not go through |
```
shreyav: this sounds like QUOTE_EXECUTION_FAILED?

openapi/components/schemas/transactions/OutgoingTransactionFailureReason.yaml
```diff
@@ -43,3 +48,13 @@ description: |
   | `COMPLIANCE_REJECTED` | The payout partner rejected the recipient on compliance grounds — sanctions, watchlist, or KYC/AML screening |
   | `LIGHTNING_PAYMENT_FAILED` | Deprecated — superseded by `QUOTE_EXECUTION_FAILED`. The Lightning settlement leg failed |
   | `COUNTERPARTY_POST_TX_FAILED` | Deprecated — coarse fallback retained for historical transactions when no granular payout reason is available |
+  | `SWEEP_BELOW_MINIMUM` | The payment settled into a rule-based account but was below the smallest amount the corridor to the rule's destination can carry |
```
shreyav: can we collapse mix/max into one

openapi/components/schemas/transactions/OutgoingTransactionFailureReason.yaml
```diff
@@ -43,3 +48,13 @@ description: |
...
+  | `SWEEP_ABOVE_MAXIMUM` | The payment settled into a rule-based account but was above the largest amount the corridor to the rule's destination can carry |
+  | `SWEEP_NO_ELIGIBLE_RAIL` | No payment rail available to the rule's destination could carry the sweep |
```
shreyav: hmm what does this mean? kinda sounds like ACCOUNT_CANNOT_RECEIVE or ACCOUNT_INVALID?

openapi/components/schemas/transactions/IncomingTransaction.yaml
```diff
@@ -28,4 +30,8 @@ allOf:
         description: Included for all transactions except those with "CREATED" status
       failureReason:
         $ref: ./IncomingTransactionFailureReason.yaml
-        description: If the transaction failed, this field provides the reason for failure.
+        description: >-
+          If the transaction failed, this field provides the reason for failure.
+          A payment that settled into a rule-based account but could not be
```
shreyav: nit i dont think this is always true now, maybe drop?

shreyav (approved):

shreyav (approved):

PR #897, opened 2026-09-04
feat(cards): add PIN management endpoints

DhruvPareek:
## Reason

Cardholders have no way to set a PIN on a Grid-issued card. Without one, PIN-verified transactions decline, and a cardholder who blocks their PIN through incorrect attempts has no recovery path through Grid at all.

## Overview

Adds three endpoints and one field. The two set paths are alternatives — a platform picks whichever matches how much of the PIN-entry experience it wants to own — and neither lets a PIN reach Grid.

| Endpoint | Purpose |
|----------|---------|
| `POST /cards/{id}/pin` | Set/replace the PIN from a PIN block the *client* encrypted for the card processor. Grid forwards the block verbatim; it is never stored, logged, or readable by Grid. For platforms with their own PIN-entry UI. |
| `POST /cards/{id}/pin/session` | Mint a short-lived session token for the processor's hosted PIN-entry iframe. The PIN crosses neither the platform's servers nor Grid's. For platforms without their own PIN UI. |
| `POST /cards/{id}/pin/unblock` | Clear a PIN blocked by three consecutive incorrect entries, leaving the PIN itself unchanged. |

`Card.pinStatus` (`NOT_SET` / `OK` / `BLOCKED`) makes the state observable, and is absent on cards whose processor offers no PIN management.

Notes for review:

- **The session response is a token, not a URL.** This is the one place the shape departs from `POST /cards/{id}/reveal`: the processor's embed-session endpoint answers with an opaque session credential and no URL, and the iframe `src` its SDK builds is not publicly documented. Returning a `pinEmbedUrl` would have meant inventing one, so the response carries `sessionToken` and the client's embed SDK builds the iframe from it.
- **`encryptedPinBlock` is deliberately opaque.** Grid cannot validate it, so the docs put the burden of correct encryption on the client and say plainly that a rejected block returns `400 INVALID_INPUT` with no finer diagnosis.
- **The iframe path is fire-and-forget.** Grid is not told whether the cardholder finished PIN entry, so `CardPinSessionResponse` points readers at `pinStatus` to confirm the outcome rather than implying the mint itself means success.
- **Unblock is a dedicated endpoint** rather than `PATCH /cards/{id}` with a status flag: `CardUpdateRequest` already requires one of `state`/`fundingSources`/`maxSpend*` and carries `CLOSED` exclusivity rules, and threading a one-way status through that is more churn than a 204.

Everything else follows the reveal endpoint — same security scheme, same audit-logging language, same short-lived-bearer-secret warnings.

## Test Plan

- `make build` — bundle regenerates cleanly.
- `npx @redocly/cli lint openapi.yaml` — valid, and the warning count is unchanged from `main` (55 before, 55 after), so these files add none.
- `npx @stoplight/spectral-cli lint openapi.yaml --fail-severity=error` — 0 errors, and no finding references any new schema or path (the info count drops by one against baseline, since `CardPinStatus` carries the example `CardState` lacks).

The sparkcore implementation and the ops UI are stacked in webdev: lightsparkdev/webdev#34390 → #34391 → #34392 → #34393 → #34394.

https://claude.ai/code/session_01GsstRe1f8aoPXm1kYQw2Vm

### DIFF
--- a/openapi/components/schemas/cards/CardPinSessionRequest.yaml
+++ b/openapi/components/schemas/cards/CardPinSessionRequest.yaml
@@ -1,0 +1,13 @@
+type: object
+required:
+  - targetOrigin
+properties:
+  targetOrigin:
+    type: string
+    format: uri
+    description: >-
+      Canonical HTTPS origin of the page that will frame the returned
+      `pinEmbedUrl`. The processor refuses to render the iframe on any other
+      origin, so this must match the embedding page's origin exactly —
+      scheme and host, no path or trailing slash.
+    example: https://app.example.com
--- a/openapi/components/schemas/cards/CardPinStatus.yaml
+++ b/openapi/components/schemas/cards/CardPinStatus.yaml
@@ -1,0 +1,15 @@
+type: string
+enum:
+  - NOT_SET
+  - OK
+  - BLOCKED
+example: OK
+description: |
+  State of the card's PIN. Absent on cards whose processor does not offer
+  PIN management.
+
+  | Status | Description |
+  |--------|-------------|
+  | `NOT_SET` | No PIN has been set on the card. PIN-verified transactions decline. Set one with `POST /cards/{id}/pin` or `POST /cards/{id}/pin/session`. |
+  | `OK` | A PIN is set and usable. |
+  | `BLOCKED` | The PIN was entered incorrectly three consecutive times and is refused until cleared with `POST /cards/{id}/pin/unblock`, or replaced with `POST /cards/{id}/pin`. |
--- a/openapi/paths/cards/cards_{id}_pin.yaml
+++ b/openapi/paths/cards/cards_{id}_pin.yaml
@@ -1,0 +1,94 @@
+parameters:
+  - name: id
+    in: path
+    description: System-generated unique card identifier
+    required: true
+    schema:
+      type: string
+post:
+  summary: Set a card PIN
+  description: >-
+    Set or replace the card's PIN from a PIN block your client encrypted for
+    the card processor. Grid forwards the block verbatim and never sees the
+    PIN: it is not stored, not logged, and not readable by Grid at any point.
+
+
+    Use this when you operate your own PIN-entry UI. To hand PIN entry to
+    the processor's hosted iframe instead, use
+    `POST /cards/{id}/pin/session` — the two are alternatives, and a card may
+    use either at any time.
+
+
+    Setting a PIN clears a `BLOCKED` PIN, so a cardholder locked out by
+    incorrect attempts can recover by choosing a new PIN without a separate
+    unblock. On success the card's `pinStatus` becomes `OK`.
+
+
+    Every call is audit-logged with the requesting actor.
+  operationId: setCardPin
+  tags:
+    - Cards
+  security:
+    - BasicAuth: []
+  requestBody:
+    required: true
+    content:
+      application/json:
+        schema:
+          $ref: ../../components/schemas/cards/CardPinSetRequest.yaml
+  responses:
+    '204':
+      description: PIN set.
+    '400':
+      description: >-
+        Bad request. Returned with `INVALID_INPUT` when `encryptedPinBlock`
+        is absent, is not valid base64, or is rejected by the card
+        processor as a malformed or mis-encrypted PIN block.
+      content:
+        application/json:
+          schema:
+            $ref: ../../components/schemas/errors/Error400.yaml
+    '401':
+      description: Unauthorized
+      content:
+        application/json:
+          schema:
+            $ref: ../../components/schemas/errors/Error401.yaml
+    '403':
+      description: >-
+        Forbidden. The session has no attributable actor to audit the change
+        against (for example, an impersonated dashboard session).
+      content:
+        application/json:
+          schema:
+            $ref: ../../components/schemas/errors/Error403.yaml
+    '404':
+      description: Card not found
+      content:
+        application/json:
+          schema:
+            $ref: ../../components/schemas/errors/Error404.yaml
+    '409':
+      description: >-
+        Conflict. Returned with `CARD_ALREADY_CLOSED` for a `CLOSED` card,
+        and with `UNSUPPORTED_CARD_OPERATION` when the card's processor does
+        not offer PIN management.
+      content:
+        application/json:
+          schema:
+            $ref: ../../components/schemas/errors/Error409.yaml
+    '500':
+      description: Internal service error
+      content:
+        application/json:
+          schema:
+            $ref: ../../components/schemas/errors/Error500.yaml
+    '501':
+      description: >-
+        Not implemented in this environment. Cards are not enabled for
+        every Grid deployment; environments without a configured card
+        issuer return `501 NOT_IMPLEMENTED`.
+      content:
+        application/json:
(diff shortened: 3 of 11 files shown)


DhruvPareek:
* **#897** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/897?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/897?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

mintlify/cards/card-management/pins.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+## Choose a PIN-entry flow
+
+For hosted PIN entry, call `POST /cards/{id}/pin/session` with the canonical HTTPS `targetOrigin` of your frontend. Include the scheme, host, and optional port, without a path, trailing slash, credentials, query, or fragment.
+
+Install the `lithic-embed` frontend SDK. Initialize `LithicEmbed` with the returned `environment` (`Environment.SANDBOX` or `Environment.PRODUCTION`), call `pinSetting(sessionToken)`, mount the control, and call its `submit()` method when the cardholder confirms. Follow [Lithic's embed integration guide](https://docs.lithic.com/docs/embedded-card-ui#setting-a-pin) for SDK usage and error handling. The control accepts four digits. The cardholder's plaintext PIN goes directly to Lithic.
+
+The token expires at `expiresAt` and permits one successful submission. Request a new session for another change or after expiration. Treat the token as a secret: never persist, cache, log, or send it to analytics. Use the SDK's submission result to confirm success. An `OK` status alone cannot prove that an already configured PIN was changed.
```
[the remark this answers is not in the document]
DhruvPareek: fixed

openapi/components/schemas/cards/CardPinSessionRequest.yaml
```diff
@@ -0,0 +1,14 @@
+type: object
+required:
+  - targetOrigin
+properties:
+  targetOrigin:
+    type: string
+    format: uri
```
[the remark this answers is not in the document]
DhruvPareek: fixed

openapi/paths/cards/cards_{id}_pin.yaml
```diff
@@ -0,0 +1,135 @@
+parameters:
```
shreyav: do we need this? seems like get card already gives you the pin

openapi/components/schemas/cards/CardPinStatus.yaml
```diff
@@ -0,0 +1,17 @@
+type: string
+enum:
+  - NOT_SET
+  - OK
+  - BLOCKED
+example: OK
+description: |
+  State of the card's PIN. On the Card resource, this is the last known status
+  and is present whenever PIN management is supported. An absent value means
+  PIN management is unavailable for this card; it does not mean `NOT_SET`.
```
shreyav: should this be a card capability

openapi/components/schemas/cards/CardPinSessionRequest.yaml
```diff
@@ -0,0 +1,15 @@
+type: object
```
shreyav: what happens if you call create session twice? should the session resource have an id?

openapi/paths/cards/cards_{id}_pin.yaml
DhruvPareek: get card only gives you the pin status, like if the pin was configured, not configured, or blocked

openapi/components/schemas/cards/CardPinStatus.yaml
DhruvPareek: that sentence is just there because Striga cards operate differently with PINs so this is just an absent value / null for striga cards

DhruvPareek: just wanted to clarify the difference between null and NOT_SET

openapi/components/schemas/cards/CardPinSessionRequest.yaml
DhruvPareek: wait one sec, turns out Ben implemented card reveal iframes using their old flow which doesnt use sessions

DhruvPareek: but why do you think a session id would be necessary? sessions are short lived and not really like an object that would be used later on or need to be referred to

openapi/paths/cards/cards_{id}_pin.yaml
shreyav: oh sry i meant the get below. seems like it only has the status

openapi/components/schemas/cards/CardPinSessionRequest.yaml
shreyav: the question underneath is what happens to session 1 when you create session 2. because the status, and the unblock api are scoped to the card and not the session

openapi/paths/cards/cards_{id}_pin.yaml
DhruvPareek: Oh yea get pin here is redundant with the existing get card. I can remove it, i'm just curious if it is something where it would seem confusing to an integrator that they cant do GET /cards/{id}/pin when we have an existing POST /cards/{id}/pin?

DhruvPareek: If i'm wrong about that, i can def remove GET /cards/{id}/pin

openapi/components/schemas/cards/CardPinSessionRequest.yaml
DhruvPareek: that's a lithic thing but they aren't explicit in their docs about it, it does Look like either session can still update the card though

DhruvPareek: so it shouldnt matter

DhruvPareek: Yea highly confident either can update the PIN still

openapi/paths/cards/cards_{id}_pin.yaml
shreyav: i dont think we need it but maybe @pengying has some thoughts. should we also emit some webhooks on these status changes?

DhruvPareek: For chaning a pin to OK or NOT_SET I don't think we need webhooks because the platform itself is making API calls to get the PIN to that state. For BLOCKED i can add follow up PRs for the webhook

openapi/paths/cards/cards_{id}_pin.yaml
```diff
@@ -0,0 +1,93 @@
...
+      type: string
+post:
+  summary: Set a card PIN
+  description: >-
+    Set or replace the card's online PIN using a block encrypted in your
+    client with the published PIN encryption public key. Grid cannot decrypt
+    the block and never stores or logs PIN material.
+
+
+    Use this when you operate your own PIN-entry UI. For a hosted PIN-entry
```
pengying: Is there a reason we wouldn't recommend using the set pin iframe? 

I feel like a pin passing through the platform services isn't PCI compliant?

PR #929, opened 2026-09-09
Document the 409 that card reveal and the sandbox simulators return

kphurley7:
## Why

Thirteen card operations return `409` in production for an entire class of card, and not one of them declared a `409` response.

- `POST /cards/{id}/reveal` declared 200, 401, 403, 404, 500, 501.
- All twelve `POST /sandbox/cards/{id}/simulate/*` operations declared 202, 400, 401, 403, 404, 500.

A generated SDK therefore has no typed branch for a status these operations really answer with, and a reader has no way to learn from the spec that a whole card program is excluded.

The reveal operation's description also implied every card can be revealed, which is what sends an integrator looking for a bug when they get the 409.

## What the code actually does

**Reveal** (`create_card_reveal.py`): the handler refuses on the card's provider kind alone, before any credential lookup, so the answer never depends on whether provider configuration happens to be present. Cards whose issuer decides authorizations have no Grid-minted PAN reveal at all; the reveal is the issuer's own challenge-based hosted flow. The already-declared `501` is a different condition: card issuance not enabled in the deployment.

**Sandbox** (`gen_require_lithic_simulable` in `sandbox_simulate_common.py`): simulated state on a provider-authoritative card would contradict the issuer's own ledger, so the guard refuses with `409`. It fires for all three ways a caller can name such a card: the canonical `Card:<uuid>`, the raw processor token `POST /cards` returns, and the issuer-side card id.

## What changed

A `409` on the reveal operation and on all twelve simulate operations, plus a corrected reveal description. Every simulate operation gets the identical description, because it is the identical condition.

Each description says it is a property of the card's program rather than a transient state, so a caller knows retrying will not help.

## Notes for review

- Twelve simulate operations, not ten. `authorization_expiry` and `authorization_reversal` are both already documented.
- The 409 descriptions deliberately do not claim that every non-simulable card gets a 409. A card belonging to **another platform** is answered as not-found instead, so that a raw provider id cannot be used to probe cross-tenant card existence. Wording that contradicted that would have been a leak in the docs.
- Response declarations only. No request schemas, examples, or descriptor-suffix outcome tables were touched.

### DIFF
--- a/mintlify/openapi.yaml
+++ b/mintlify/openapi.yaml
@@ -8874,7 +8874,9 @@ paths:
     post:
       summary: Reveal card details
       description: |-
-        Mint a signed, short-lived URL for the card processor's iframe that displays the card's full PAN, CVV, and expiry to the cardholder. This is the only way to obtain a reveal URL — the `Card` resource never carries one.
+        For card programs where Grid makes authorization decisions, mint a signed, short-lived URL for the card processor's iframe that displays the card's full PAN, CVV, and expiry to the cardholder. The `Card` resource never carries a reveal URL.
+
+        Cards in programs where the card issuer makes authorization decisions use the card issuer's challenge-based hosted reveal flow instead and cannot be revealed through this endpoint.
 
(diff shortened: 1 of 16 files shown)


kphurley7:
@faraday review

kphurley7:
@greptileai review

jklein24 (approved):

PR #931, opened 2026-09-10
feat(rails): add ACH_SAME_DAY payment rail

jacklatourette:
## What

Adds `ACH_SAME_DAY` to the `PaymentRail` enum alongside the existing `ACH`, and to the
`paymentRails` array on `UsdAccountInfo`.

## Why

Grid exposes one ACH rail today. Every ACH payout it originates already settles same
day, because the sponsor bank's default service level is same-day and nothing in the
request path ever asks for a different one. A platform therefore cannot choose
standard ACH, and the two speeds cannot be priced apart.

Splitting the rail gives platforms an explicit choice, and gives each speed its own
`RailFeeConfig`.

## The `paymentRails` array is the part that matters at runtime

The top-level enum alone is not enough. The generated client validates an account's
`paymentRails` against the array on `UsdAccountInfo`, so advertising the rail on a USD
account without that entry makes `GET /customers/external-accounts/{id}` return **500**
— verified against a live account.

`UsdAccountInfo` is composed by `PaymentUsdAccountInfo` and `UsdExternalAccountInfo`, so
the single entry covers all three generated models.

## What this means for integrators

- `ACH_SAME_DAY` requests same-business-day settlement and is priced separately. Above
  the NACHA per-entry same-day limit it settles on the standard schedule and is priced
  accordingly — it is **not** rejected.
- `ACH` **will** settle on the standard next-business-day schedule. Until an announced
  date it continues to settle same-business-day on production platforms, and already
  settles next-business-day on sandbox platforms. The descriptions say so explicitly
  rather than claiming next-day everywhere, because that would be wrong for every
  production customer until the cutover.
- `RailFeeConfig` accepts `ACH_SAME_DAY`. The two are separate rails, so price each one
  with its own config.

The change is additive to both enums. No existing field changes shape.

## Docs

The pages that enumerate rails now list it: the US row of the country support table, the
`paymentRail` tip on both sending pages, and the rail-fee limits on the fees page. The
changelog gets a September 2026 entry.

## Sequencing

Backend support lands separately in webdev under AT-6555, which regenerates its vendored
client from this branch. Until that ships, a request naming `ACH_SAME_DAY` is declined by
the backend, so hold this if you would rather the docs never advertise a rail the API
does not yet accept.

## Verification

- Rebased onto current `main` (was 45 behind). One conflict, in the changelog, where both
  sides added September entries; both kept.
- `make build` rebundles cleanly; `openapi.yaml` and `mintlify/openapi.yaml` are in sync.
- `make lint-openapi` exits 0: **866 problems, 0 errors** (174 warnings, 692 infos, all
  pre-existing and unrelated — the 2 redocly warnings are `paths-kebab-case`).
- `make lint-markdown` cannot run locally: `package.json` has no `lint:markdown` script.
  CI's "Lint Code & Documentation" job covers it.

https://claude.ai/code/session_01RUF3krYG9C9hShf4iS5Sug


jacklatourette:
* **#931** <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/931?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/lightsparkdev/grid-api/931?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

bsiaotickchong (approved):

ch-brian (approved):

PR #940, opened 2026-09-10
Document that fundingSources cannot be combined with a state change

kphurley7:
## Why

A companion webdev PR adds a guard to `PATCH /cards/{id}`: on card programs where the card issuer makes authorization decisions, `fundingSources` may no longer be supplied in the same request as a `state` change. This documents that rule.

The behaviour it prevents is a durable half-apply reported as a failure.

On those programs the funding change is applied at the issuer and **committed** before the state change is attempted, and nothing can undo it if the state change then fails. A caller sending `{fundingSources: [B], state: "FROZEN"}` could receive an error for a request that had already, permanently, changed where the card draws its money from - while leaving the card unfrozen.

That is the worst half to survive: the leg that would have stopped the card spending is the one that failed, and the leg that redirected its funding is the one that stuck. Refusing the combination up front is the only way to keep the request all-or-nothing.

## What changed

- The `fundingSources` bullet in the `patch` description now states the restriction, and that it does **not** apply where Grid makes the authorization decision (those two fields are applied atomically there, so the combination stays valid).
- The `400` description names the new cause.

## Notes for review

- The restriction is **broader** than the `state: CLOSED` clause it sits beside: that one is about a single target state, this one is about any state change, on a subset of card programs. The bullet distinguishes the two rather than merging them.
- Wording follows the phrasing already used across this file and the reveal operation - "card programs where the card issuer makes authorization decisions" / "where Grid makes the authorization decision". An earlier draft stated the condition negatively; that introduced a third form for a distinction that now appears in about a dozen places, so it was rewritten positively to match.
- Deliberately untouched: `customerId`, `stateReason`, the limit fields, and `FUNDING_SOURCE_INELIGIBLE`. Another team has active work in this file.
- `info.version` is not bumped: this documents a validation rule and the companion PR carries the behaviour.

**Merge after webdev #34760**, or this describes a rule that is not yet enforced.


kphurley7:
@faraday review

kphurley7:
@greptileai review

jklein24 (approved):

PR #943, opened 2026-09-10
Add card capabilities so conditional fields become discoverable

kphurley7:
## Why

The card surface makes behaviour conditional on the card program in **ten** places, and nothing in the API tells a caller which program applies. There is no field on `Card`, none on `InternalAccount`, and no up-front error that names it. A caller discovers the answer by sending a request and getting a `400`.

That is tolerable for an optional feature. It is not tolerable when it decides whether a request can be constructed at all, and a platform can run both programs at once.

## Why the discriminator is a capability, not the authorization authority

This is the design constraint, and it is already demonstrable rather than theoretical.

In Grid, the three limit fields are gated **differently**:

- `maxTransactionsPerDay` is refused when the provider owns transaction authority. Authority-based.
- `maxSpendPerTransaction` / `maxSpendPerDay` are refused only when the provider does not satisfy a `LimitsCapable` protocol. Capability-based.

The EU issuer now satisfies `LimitsCapable` - Grid forwards spend limits to it - so **those two fields work today on cards whose issuer makes authorization decisions**, while the spec still says they are "Supported only for card programs whose authorization decisions are made by Grid."

That sentence is already false, and it went false without anyone editing the spec. An `authority` enum would need a coordinated spec change every time a capability and the authority drift apart, and they have already drifted once inside one release cycle.

## Why not the card currency

Routing resolves the account owner's **pinned switch first**, falling back to a currency-wide lookup only when there is no pin. Two customers on one platform with the same currency can be routed to different issuers if they onboarded differently.

Separately, the EU issuer is not euro-only: its card-currency set is deliberately unconstrained and its default spending set spans EUR plus several crypto currencies. A currency discriminator would be right most of the time, which is worse than none, because clients encode it.

## What changed

`CardCapabilities` - four read-only booleans naming what a caller can **do**: `supportsSpendLimits`, `supportsTransactionCountLimit`, `supports3dSecurePassword`, `supportsPanReveal`.

Surfaced in two places, because the question is asked at two different times:

- **`InternalAccount.cardCapabilities`** - the pre-issuance answer. `fundingSources` is required on `POST /cards`, so a caller always holds the account before they can build the request.
- **`Card.cardCapabilities`** - the post-issuance answer, required, since a card always has an issuer.

All ten conditional sentences now point at the flag that answers them, plus an eleventh site found while doing it: the reveal endpoint's `409` description.

## The two placements are not mirrors

This is the subtlety worth reviewing carefully. An earlier draft of this change treated the card's copy as an echo of the account's. It is not.

A card's issuer is selected from the funding account at issuance and then **persisted on the card**; everything afterwards resolves from the card, not from its current funding source. But a platform can be re-routed later - the implementation explicitly handles "a platform that has since re-routed EUR away from Striga". When that happens the account starts predicting a different issuer while existing cards keep the one they were issued under.

So the two can legitimately disagree, permanently. The descriptions say so: the account carries a **prediction** that can change, the card carries a **fact** fixed at issuance.

## One behaviour this documents for the first time

`POST /cards` accepts an array of funding sources and resolves the issuer from **`funding_sources[0]` only**. The spec already gives that array's order a different meaning - "Array order determines the priority Authorization Decisioning tries them in" - so a caller reordering it for spend-priority reasons could change **who issues the card**, with nothing telling them. That is existing behaviour, now stated.

## Scope

Additive read-only fields plus description rewrites. `customerId`, `fundingSources` as a field, `stateReason` and the limit fields are untouched; another team has active work there. `CardConfig` and the sandbox simulator responses also carry authority language and were deliberately left alone, because their behaviour is not represented by these four flags.

`info.version` is not bumped.

**This is a design proposal as much as a change** - it is the concrete version of the discussion about how to make these conditions discoverable, and the flag set is the part most worth arguing with.

### DIFF
--- a/openapi/components/schemas/cards/Card.yaml
+++ b/openapi/components/schemas/cards/Card.yaml
@@ -61,6 +61,12 @@ properties:
     example:
       - InternalAccount:019542f5-b3e7-1d02-0000-000000000002
       - InternalAccount:019542f5-b3e7-1d02-0000-000000000003
+  cardCapabilities:
+    $ref: ./CardCapabilities.yaml
+    description: >-
+      Actions supported for this card by the issuer selected at issuance.
+      Present for cards whose program has been resolved; absent otherwise.
+      These capabilities are fixed at issuance for the card's lifetime.
   maxSpendPerTransaction:
     anyOf:
       - type: integer
--- a/openapi/components/schemas/cards/CardCapabilities.yaml
+++ b/openapi/components/schemas/cards/CardCapabilities.yaml
@@ -1,0 +1,35 @@
+type: object
+description: Actions supported by the card program associated with this resource.
+required:
+  - supportsSpendLimits
+  - supportsTransactionCountLimit
+  - supports3dSecurePassword
+  - supportsPanReveal
+properties:
+  supportsSpendLimits:
+    type: boolean
+    description: >-
+      Whether cards in this program accept `maxSpendPerTransaction` and
+      `maxSpendPerDay`.
+    readOnly: true
+    example: true
+  supportsTransactionCountLimit:
+    type: boolean
+    description: >-
+      Whether cards in this program accept `maxTransactionsPerDay`.
+    readOnly: true
+    example: true
+  supports3dSecurePassword:
+    type: boolean
+    description: >-
+      Whether cards in this program accept a caller-supplied
+      `threeDSecurePassword`.
+    readOnly: true
+    example: false
+  supportsPanReveal:
+    type: boolean
+    description: >-
+      Whether cards in this program can be revealed through
+      `POST /cards/{id}/reveal`.
+    readOnly: true
+    example: true
--- a/openapi/components/schemas/cards/CardCreateRequest.yaml
+++ b/openapi/components/schemas/cards/CardCreateRequest.yaml
@@ -20,12 +20,15 @@ properties:
   threeDSecurePassword:
     type: string
     description: >-
-      Optional static password used as the card's 3-D Secure factor. Only
-      accepted for card programs whose issuer supports a static-password
-      factor (EU cards today); supplying it for a program that does not is
-      rejected with `INVALID_INPUT`. When omitted, one is generated on the
-      cardholder's behalf. Grid does not retain the value: it is forwarded to
-      the issuer and discarded, so it cannot be read back afterwards.
+      Static password used as the card's 3-D Secure factor. Required when the
+      first funding-source internal account's
+      `cardCapabilities.supports3dSecurePassword` is true; omitting it or
+      supplying an empty or whitespace-only string is rejected with
+      `INVALID_INPUT`. When the capability is false, supplying this field is
+      rejected with `INVALID_INPUT` because cards in that program have no
+      static-password factor. Grid does not retain the value: it is forwarded
+      to the issuer and discarded, so it cannot be read back afterwards; a
+      cardholder who forgets it must set a new one through `PATCH /cards/{id}`.
     example: AbCd1234EfGh5678
   form:
     $ref: ./CardForm.yaml
@@ -33,6 +36,8 @@ properties:
     type: array
     description: >-
       Internal account ids to bind as funding sources, in priority order.
+      The first entry selects the card issuer and therefore the card's
+      capabilities.
       The first entry is tried first by Authorization Decisioning. Every
       card must be bound to at least one source, and every source must
       belong to the cardholder and be denominated in a card-eligible
@@ -53,9 +58,9 @@ properties:
       of the card currency derived from its funding sources. Omit this field
       for no card-specific cap. When the platform config also supplies
       `cardConfigs.maxSpendPerTransaction`, Grid enforces the lower of the two
-      values. Supported only for card programs whose authorization decisions
-      are made by Grid. A transaction for exactly the effective limit is
-      allowed.
+      values. Accepted only when the funding-source internal account's
+      `cardCapabilities.supportsSpendLimits` is true. A transaction for
+      exactly the effective limit is allowed.
     example: 5000
   maxSpendPerDay:
     type: integer
@@ -69,8 +74,9 @@ properties:
       platform config also supplies `cardConfigs.maxSpendPerDay`, Grid enforces
       the lower of the two values. The window resets at 00:00 UTC, and refunds,
       reversals, and authorization expiries do not restore capacity during the
-      day. Supported only for card programs whose authorization decisions are
-      made by Grid. Spend exactly equal to the effective limit is allowed.
+      day. Accepted only when the funding-source internal account's
+      `cardCapabilities.supportsSpendLimits` is true. Spend exactly equal to
+      the effective limit is allowed.
     example: 25000
   maxTransactionsPerDay:
     type: integer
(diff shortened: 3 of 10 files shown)


kphurley7:
@faraday review

kphurley7:
@greptileai review

openapi/components/schemas/customers/InternalAccount.yaml
```diff
@@ -51,6 +51,16 @@ properties:
       Whether wallet privacy is enabled for the Embedded Wallet. Only present
       for `EMBEDDED_WALLET` internal accounts.
     example: true
+  cardCapabilities:
+    $ref: ../cards/CardCapabilities.yaml
+    description: >-
+      Prediction of the actions supported for a card issued now with this
```
pengying: Do we need prediction here?  Or is it an accurate representation of the card capabilities issued against this account?

openapi/components/schemas/cards/Card.yaml
```diff
@@ -61,6 +62,13 @@ properties:
     example:
       - InternalAccount:019542f5-b3e7-1d02-0000-000000000002
       - InternalAccount:019542f5-b3e7-1d02-0000-000000000003
+  cardCapabilities:
+    $ref: ./CardCapabilities.yaml
+    description: >-
+      Actions supported for this card by the issuer selected at issuance. They
+      are stable for the card's lifetime, even if the platform's card routing
```
pengying: do we need the even if the platform's card routing changes, and do not predict the capbilities of a new card issued using the same funding sources?

openapi/components/schemas/cards/CardCreateRequest.yaml
```diff
@@ -20,9 +20,9 @@ properties:
...
     description: >-
-      Optional static password used as the card's 3-D Secure factor. Only
-      accepted for card programs whose issuer supports a static-password
-      factor (EU cards today); supplying it for a program that does not is
+      Optional static password used as the card's 3-D Secure factor. Accepted
+      only when the funding-source internal account's
+      `cardCapabilities.supports3dSecurePassword` is true; otherwise it is
       rejected with `INVALID_INPUT`. When omitted, one is generated on the
       cardholder's behalf. Grid does not retain the value: it is forwarded to
       the issuer and discarded, so it cannot be read back afterwards.
```
pengying: Are there implications of a password being generated on the cardholder's behalf?  Does that mean when they're prompted for 3DS they'll need to enter a password they never set?

pengying: I guess platforms should aways have them create a password on issuance

pengying (changes requested):
Few nits but lgtm

openapi/components/schemas/cards/Card.yaml
```diff
@@ -5,6 +5,7 @@ required:
   - state
   - form
   - fundingSources
+  - cardCapabilities
```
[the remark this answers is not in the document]
kphurley7: Valid, and sharper than stated. **Nothing populates this field yet** - the implementation side is deliberately unbuilt while the flag set is still under discussion. Merging it as `required` would mean every `GET /cards` response the server actually sends violates its own schema, and regenerated SDKs would treat a permanently-absent field as non-nullable.

Removed from `Card.required`; the property stays.

That also removed the only reason `card-state-change.yaml` carried it: the field had been added to those webhook examples purely to satisfy the required constraint, so it is out of the webhook payloads too. A webhook shouldn't carry a field just because a schema forced it.

openapi/components/schemas/cards/CardCapabilities.yaml
```diff
@@ -0,0 +1,36 @@
...
+  - supportsSpendLimits
+  - supportsTransactionCountLimit
+  - supports3dSecurePassword
+  - supportsPanReveal
+properties:
+  supportsSpendLimits:
+    type: boolean
+    description: >-
+      Whether you can supply `maxSpendPerTransaction` and `maxSpendPerDay`
+      when creating or updating a card.
```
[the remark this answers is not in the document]
kphurley7: Valid. "Whether you can supply these **when creating or updating a card**" reads correctly on `InternalAccount` and half-wrong on `Card`, where creation is already past.

Fixed by wording rather than structure. Each flag now describes the **card program** - "whether cards in this program accept X" - and the placement descriptions carry the rest: the account's copy is a prediction for a card issued now, the card's is a fact fixed at issuance. That keeps one schema and puts each statement where it is unambiguously true.

Applied to all four flags, not only the two named.

openapi/paths/cards/cards_{id}_reveal.yaml
```diff
@@ -8,15 +8,14 @@ parameters:
...
+    signed, short-lived URL for the card processor's iframe that displays the
+    card's full PAN, CVV, and expiry to the cardholder. The `Card` resource
+    never carries a reveal URL.
 
 
-    Cards in programs where the card issuer makes authorization decisions use
-    the card issuer's challenge-based hosted reveal flow instead and cannot be
-    revealed through this endpoint.
+    When `cardCapabilities.supportsPanReveal` is false, this endpoint cannot
+    reveal the card.
```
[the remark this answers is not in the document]
kphurley7: Valid, and it was a regression this PR introduced. That sentence was added deliberately in #929, which merged two days ago, precisely so a caller who gets the refusal knows what to do next. The rewrite replaced it rather than augmenting it.

Restored in both the operation description and the `409`. A caller now learns both how to predict the refusal (`supportsPanReveal`) and what to do when it happens. Telling someone how to anticipate a failure is not a substitute for telling them the way through it.

openapi/components/schemas/cards/CardCreateRequest.yaml
kphurley7: Yes. If it's omitted we generate 16 random characters, send them to Striga, and throw them away (`striga_grid_switch.py:4856`). The card ends up enrolled with a static password nobody can produce, so the cardholder can't pass 3DS.

#34601 added rotation so one can be set afterwards, but that only helps if someone notices.

You're right that platforms should always set it at issuance. Making it required is free right now since no Striga cards have been issued yet, so nothing to break. Fixing it.

openapi/components/schemas/customers/InternalAccount.yaml
kphurley7: Fair - "prediction" is the wrong word and I chose it badly.

It is accurate, not probabilistic: at the moment you read it, this account routes to a specific issuer and those are that issuer's capabilities. A card issued right now gets exactly them.

The two things I was trying to convey are about **time**, not accuracy:

- it can change later, if the platform's card routing changes
- it does not describe cards already issued from this account, whose capabilities were fixed when they were issued

Those matter because the two can diverge permanently - re-route a platform and its accounts start pointing at a different issuer while its existing cards keep the old one - but "prediction" implies uncertainty about the present, which is wrong. Rewording to say it plainly: accurate for a card issued now, subject to change, not a statement about existing cards.

openapi/components/schemas/cards/Card.yaml
kphurley7: No, probably not both. That is over-explained and I wrote it that way because I had just worked out the divergence case and over-corrected.

The load-bearing fact is that these are **fixed at issuance**. Someone who understands that can derive the rest; routing changing later is a scenario most readers will never meet, and the cross-reference to "a new card issued using the same funding sources" belongs on the account field, not here, if anywhere.

Trimming to the fixed-at-issuance statement.

openapi/components/schemas/cards/CardCreateRequest.yaml
kphurley7: Fixed in webdev #34868. Issuance now refuses a missing, empty, or whitespace-only password instead of generating one, and the generator is deleted so it can't come back.

Left the spec wording here for now since it's a separate question - the field is genuinely unsupported on Lithic, so it can't just move to `required` on the schema. Probably wants a sentence saying it's required where `supports3dSecurePassword` is true.

pengying (approved):

PR #960, opened 2026-09-12
feat(cards): report why a card authorization was declined

whoisglover:
Jira: [ENG-11677](https://lightspark.atlassian.net/browse/ENG-11677) under [ENG-11645](https://lightspark.atlassian.net/browse/ENG-11645)

## What this does

When a card authorization is declined, the card transaction reports `status: DECLINED`. #957 added that status.

A platform can now see that the authorization failed, but not why. A frozen card, a spend cap, and an empty funding source all look the same. Only some of those are something the platform can act on, and it cannot tell them apart.

This PR adds the reason.

## How it works

- New `CardDeclinedReason` enum: `CARD_NOT_ACTIVE`, `SPEND_LIMIT_EXCEEDED`, `INSUFFICIENT_FUNDS`, `NO_ELIGIBLE_FUNDING_SOURCE`, `BLOCKED`, `UNSUPPORTED_NETWORK`, `OTHER`.
- `CardTransaction` gains `cardDeclinedReason`. It is present only when `status` is `DECLINED`.
- The `CARD_TRANSACTION.DECLINED` webhook example carries the reason.
- Sandbox descriptor suffix `003` documents the reason it produces, which is `CARD_NOT_ACTIVE`.

It also retires a name that was never real. The docs told platforms a frozen card declines with `CARD_PAUSED`. No enum has ever had that member. The server's decision vocabulary calls it `CARD_NOT_ACTIVE`. Every page that said `CARD_PAUSED` now says `cardDeclinedReason: CARD_NOT_ACTIVE`.

## What has to land with this

The server does not expose the reason yet. It stores it on the decision row, `EntCardEventDecision.decline_reason`, and the public projection drops it. The webdev change that joins the decline decision to the transaction and emits the field merges in the same window.

## One thing to settle before this merges

`INSUFFICIENT_FUNDS` is in the enum, and nothing produces it.

Four of the seven reasons are real in production today. `NO_ELIGIBLE_FUNDING_SOURCE`, `SPEND_LIMIT_EXCEEDED`, `BLOCKED`, and `UNSUPPORTED_NETWORK` all come from the shared gates that run after `_gen_evaluate` in `decision/base.py`, so they apply on production platforms even though `ProductionCardDecisionEngine` itself approves everything.

`INSUFFICIENT_FUNDS` is not one of them. Grid never checks the funding source balance when it decides an authorization. The auth is approved to the card network, and the pull fails afterwards in `card_hold_action.py`, whose own comment says so: "The auth was already approved to the issuer; if only the pull fails (e.g. the wallet is short) the raw exception propagates as transient so the whole delivery rolls back and Lithic re-delivers." That path resolves as `EXCEPTION`, and it carries no decline reason.

So the docs in this PR now say that plainly, on the sandbox `002` row and on the pre-funding guide. What is left to decide is the enum itself. Keeping the member means committing to move the balance check to authorization time, which `production.py` already has a TODO for. Dropping it means adding it back later, which is breaking for strict clients on a response enum.

`CARD_NOT_ACTIVE` has a smaller version of the same question: only `SandboxDecisionEngine` checks card status, and no shared gate does, so on production platforms a frozen card is declined by the provider rather than by us.

## Tests

`make build` rebundles cleanly. `make lint` reports 0 errors. The one new finding is informational and is the known `$ref`-with-example shape that `direction` and `applicationFee` already report. `cd cli && npm test` passes 80 tests. No `CARD_PAUSED` remains anywhere under `openapi/` or `mintlify/snippets/`.

## History

This is the surviving half of #935. That PR was merged into #934's branch by mistake rather than into `main`, and #934 has since been rebuilt on `main` without it.

The other half of #935 was the `DECLINED` status itself, which #957 shipped independently. ENG-11676 is therefore already delivered and only the reason was left to land.

[ENG-11677]: https://lightspark.atlassian.net/browse/ENG-11677?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ENG-11645]: https://lightspark.atlassian.net/browse/ENG-11645?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mintlify/openapi.yaml
+++ b/mintlify/openapi.yaml
@@ -8799,7 +8799,7 @@ paths:
         This endpoint is authenticated by the platform credential alone and returns `200` directly. It deliberately does not use Grid's 202 → signed-retry pattern: that pattern signs with the session key of a credential on the owning internal account, so it models actions taken *by* the end user on their own credentials or funds. Freezing or closing a card is routinely an action taken *about* a user and without them present - fraud response, offboarding, an ops-driven freeze - and requiring the cardholder's signature would make exactly those cases impossible. Operations that expose sensitive card data (`POST /cards/{id}/reveal`, 3DS password retrieval) are SCA-railed instead, because there the cardholder is the party being served.
 
         Effects:
-        - `state: FROZEN`: Authorization Decisioning declines new auths with `CARD_PAUSED`. Existing pulls and in-flight reconciliation continue — freezing does not pause the lifecycle of authorizations that already passed.
+        - `state: FROZEN`: Authorization Decisioning declines new auths with `cardDeclinedReason: CARD_NOT_ACTIVE`. Existing pulls and in-flight reconciliation continue — freezing does not pause the lifecycle of authorizations that already passed.
         - `state: ACTIVE`: normal authorization behavior resumes.
         - `state: CLOSED`: terminal close. The card transitions to `state: "CLOSED"` with `stateReason: "CLOSED_BY_PLATFORM"` and stays in the system for audit and reconciliation. All pending auths reconcile to a terminal state via the existing reconcile primitive. Inbound clearings received after close follow the standard force-post / late-presentment path — Lightspark absorbs the loss if a post-hoc pull on the now-unbound source fails. Funding-source bindings are detached. Refunds already in flight still complete because Lightspark holds the card-reserve keys.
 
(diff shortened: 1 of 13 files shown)


whoisglover:
@greptile review

whoisglover:
@greptile review

PR #968, opened 2026-09-12
docs(cards): carry the merchant descriptor in a card transaction's description

whoisglover:
https://lightspark.atlassian.net/browse/ENG-11675

## What this does

`GET /transactions` returns a mixed list of transactions. Every type in that list carries a `description` field that says, in one line, what the money was for.

Card transactions are the exception. Their merchant name only lives inside the `merchant` object, so a platform rendering a transaction list has to open each card row to find out where the money went.

This adds `description` to the card transaction schema. It carries the merchant descriptor, the same string as `merchant.descriptor`.

## How it works

`description` on a card transaction is set by Grid from the descriptor the card network sends. On other transaction types the platform supplies it. The schema says so, because a reader coming from the base `Transaction` will assume they can write it.

`merchant` does not change. It keeps the structured fields, including the descriptor, the merchant category code, and the country.

The field is optional, so this is not a breaking change for anyone reading the spec today.

## Follow-up

The server does not fill this field yet. That is a webdev change and it lands after this spec merges, because it needs a regenerated client.

### DIFF
--- a/mintlify/openapi.yaml
+++ b/mintlify/openapi.yaml
@@ -11921,6 +11921,7 @@ webhooks:
                     issuerTransactionToken: lithic_txn_b81c2a4f
                     status: AUTHORIZED
                     direction: DEBIT
+                    description: BLUE BOTTLE COFFEE SF
                     merchant:
                       descriptor: BLUE BOTTLE COFFEE SF
                       mcc: '5814'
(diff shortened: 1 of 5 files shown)


whoisglover:
@greptile review

ch-brian (approved):

PR #972, opened 2026-09-13
feat(cards): report why a card authorization was declined

whoisglover:
https://lightspark.atlassian.net/browse/ENG-11677

## What this does

When Grid declines a card authorization, the card transaction comes back with `status: DECLINED`. A platform can see that the authorization failed, but not why. A frozen card, a spend cap, and a card with no usable funding source all look the same, and only some of those are something the platform can act on.

The server already returns the reason. It has been returning `cardDeclinedReason` on every declined card transaction since webdev #34974 merged on 2026-09-12. The field is missing from this spec, so the public docs describe an API that does not match what the API returns.

This adds it.

## How it works

`CardDeclinedReason` is a new enum with seven values: `CARD_NOT_ACTIVE`, `SPEND_LIMIT_EXCEEDED`, `INSUFFICIENT_FUNDS`, `NO_ELIGIBLE_FUNDING_SOURCE`, `BLOCKED`, `UNSUPPORTED_NETWORK`, and `OTHER`. They match the values the server maps to, one for one.

`CardTransaction` gains an optional `cardDeclinedReason`, present only when `status` is `DECLINED`. The declined webhook example now carries it.

## Two corrections that came out of checking the server

`CARD_PAUSED` appeared in six places across the spec and the cards pages, as the reason a frozen card declines. It is not a value the API returns. The public value is `CARD_NOT_ACTIVE`, and every one of those six places now says so.

`INSUFFICIENT_FUNDS` is returned in Sandbox only. Production does not check the funding source balance when it decides an authorization, so an underfunded card is approved to the network and the pull fails afterwards, resolving the transaction as `EXCEPTION`. The enum description and both copies of the Sandbox suffix table say this now, and the cardholder setup page no longer tells platforms their cards decline at auth time when the balance is short.

## Why now

`cardDeclinedReason` reaching main here also unblocks webdev. The vendored client in that repo carries the field today even though no grid-api commit has ever had it, so a clean run of `update_schema.sh` against main would delete a field the server emits. After this merges, that regen is safe.


whoisglover:
@greptile review

AaryamanBhute (approved):

whoisglover:
### Merge activity

* **Sep 13, 8:24 AM UTC**: @whoisglover merged this pull request with [Graphite](https://app.graphite.com/github/pr/lightsparkdev/grid-api/972).

PR #985, opened 2026-09-14
add the sandbox card dispute simulators to the spec

AaryamanBhute:
## Summary

Grid's sandbox drives a card dispute from open to close, but none of the four paths were in the spec, so generated clients could not call them and integrators had no way to discover them.

Four new paths, all tagged `Sandbox` and `404` in production, like the sibling card simulators:

- `POST /sandbox/cards/{id}/simulate/dispute_open` opens a dispute against a settled `CardTransaction`. `amount` defaults to the settled amount and cannot exceed it; a transaction that already has an open dispute returns `409`.
- `POST /sandbox/cards/{id}/simulate/dispute_provisional_credit` credits the cardholder on the open dispute, paid from the card's funding source when the platform is enabled for dispute funding.
- `POST /sandbox/cards/{id}/simulate/dispute_provisional_credit_reversal` takes that credit back.
- `POST /sandbox/cards/{id}/simulate/dispute_close` closes the dispute with a disposition of `WON`, `PARTIALLY_WON`, `LOST`, `WITHDRAWN`, or `DENIED`. `PARTIALLY_WON` is the one outcome that takes a `recoveredAmount`, which must be strictly between 0 and the disputed amount and defaults to half.

Three request schemas come with them: `SandboxCardDisputeOpenRequest`, `SandboxCardDisputeCreditRequest` (shared by the two provisional-credit paths), and `SandboxCardDisputeCloseRequest`. All four paths return `SandboxCardDisputeResponse`, which reports the resulting dispute rather than only a correlating token, as the card-event simulators do: `disputeId`, `disputeToken`, `status` (`OPEN` or `CLOSED`), and `disposition` (the five close outcomes, or null while the dispute is open). Both lifecycle fields are enum-constrained. The cardholder money a dispute event moves still settles asynchronously.

The snake_case path segments mirror the deployed routes, as the sibling simulators do, so they join the Spectral and Redocly kebab-case exemption lists.

All four operations carry `x-hidden: true`, so they stay out of the published API reference while the endpoints are used internally for testing. Mintlify reads `x-hidden` at the operation level; Redocly's own mechanism is `x-internal` plus the `remove-x-internal` decorator, which this repo does not configure, so the extension passes through `make build` into both `openapi.yaml` and `mintlify/openapi.yaml` untouched. It does not affect the generated clients, which are built from the paths themselves.

One caveat for review: `mintlify/docs.json` points the API reference at the Stainless documented-spec URL rather than this repo's bundle, so whether these operations actually disappear from the rendered docs depends on Stainless preserving the extension in its output. That is verifiable on the docs preview once this merges. If Stainless drops unknown vendor extensions, the equivalent knobs are `unspecified_endpoints`, `skip`, or `document: false` in `.stainless/stainless.yml`. Hiding is a documentation concern only: the endpoints stay callable and remain guarded by the sandbox-only check that returns `404` in production.

## Changes: 13 files

- 4 path files under `openapi/paths/sandbox/cards/`, wired into `openapi/openapi.yaml`, each marked `x-hidden: true`
- 4 schema files under `openapi/components/schemas/cards/`
- `.spectral.yaml` and `.redocly.lint-ignore.yaml` — kebab-case exemptions for the four segments
- `openapi.yaml` and `mintlify/openapi.yaml` — rebundled with `make build`

## Test plan

- `make build` rebundles cleanly; the two bundles are byte-identical.
- `make lint` — 0 errors, and the same 883-problem (177 warning, 706 info) baseline as `main`. Redocly warnings stay at 2, unchanged, with the four new paths added to the ignore file alongside the existing simulator paths.
- Each path's request and response schemas round-trip through the bundled spec against the request bodies the deployed handlers accept: an open with and without `amount`, a provisional credit and its reversal, and a close as both `WON` and `PARTIALLY_WON`.
- Verified `x-hidden: true` is present on all four operations in both committed bundles, and on none of the 12 existing `/simulate/` paths.

Requested by @aaryamanbhute

Original PR: https://github.com/lightsparkdev/grid-api/pull/984


AaryamanBhute:
@bolt I will be using this internally to test, add the x-hidden property to hide it from open api spec

AaryamanBhute:
@bolt can you confirm it doesn't show in docs

JasonCWang (approved):

AaryamanBhute:
### Merge activity

* **Sep 15, 6:48 PM UTC**: @AaryamanBhute merged this pull request with [Graphite](https://app.graphite.com/github/pr/lightsparkdev/grid-api/985).

PR #679, opened 2026-07-10
Revert info.version to 2025-10-13 and document version-bump rule

shreyav:
## Summary
- Revert the Grid API OpenAPI `info.version` from `2026-06-29` back to `2025-10-13`, the value it held on the initial schema commit. Edited the source `openapi/openapi.yaml` and rebundled (`make build`), so the generated `openapi.yaml` and `mintlify/openapi.yaml` are updated in lockstep.
- Add a Critical Rules bullet to `CLAUDE.md` establishing the versioning convention: **bump `info.version` only when a breaking change ships, at which point it should match a new `servers.url` path**.

## Why
The two prior bumps ([#617](https://github.com/lightsparkdev/grid-api/pull/617): `2025-10-13` → `2026-06-25`, and [#633](https://github.com/lightsparkdev/grid-api/pull/633): `2026-06-25` → `2026-06-29`) were both purely additive, non-breaking schema changes. Under the new convention the version should not have moved for those, so it's reset to `2025-10-13` — which matches the current `servers.url` path `https://api.lightspark.com/grid/2025-10-13`.

## Test plan
- `make lint` passes (0 errors).
- Verified `info.version` reads `2025-10-13` in all three files (`openapi/openapi.yaml`, `openapi.yaml`, `mintlify/openapi.yaml`).

Requested by @shreyav

Slack thread: https://lightsparkgroup.slack.com/archives/D0AH5T8FMGE/p1783708971530589

---
🤖 [custodial-circuit](https://zeus.dev.dev.sparkinfra.net/#/arc?id=custodial-circuit)[(#1)](https://zeus.dev.dev.sparkinfra.net/#/instance?id=custodial-circuit) | [Feedback](https://zeus.dev.dev.sparkinfra.net/feedback)

Original PR: https://github.com/lightsparkdev/grid-api/pull/678

PR #990, opened 2026-09-14
feat(customers): accept typed agreement consents for seven documents

bsiaotickchong:
## Summary

- Replaces the single `endUserTermsConsent` field with `agreementConsents`, a list carrying one entry per agreement, so each document's acceptance is recorded and auditable separately.
- Adds an `AgreementType` enum with the seven supported documents, and changes `GET /customers/end-user-terms` to return a catalog of all seven as `{ data: [{ type, version, url }] }`.
- Spec + docs only. Backend, migration, and generated-client work follow in separate PRs.

## The seven agreement types

| Document | `AgreementType` |
| --- | --- |
| Lightspark Terms of Service (End User Terms) | `LIGHTSPARK_END_USER_TERMS` |
| Lightspark E-SIGN consent | `LIGHTSPARK_E_SIGN_CONSENT` |
| Lightspark Privacy Policy | `LIGHTSPARK_PRIVACY_POLICY` |
| Lead E-SIGN consent | `LEAD_E_SIGN_CONSENT` |
| Lead Accountholder Agreement | `LEAD_ACCOUNTHOLDER_AGREEMENT` |
| Lead Privacy Policy | `LEAD_PRIVACY_POLICY` |
| Lead Cardholder Agreement | `LEAD_CARDHOLDER_AGREEMENT` |

Each consent entry carries `type` plus the four existing evidence fields (`acceptedAt`, `ipAddress`, `termsVersion`, `acceptanceMethod`), all required. `termsVersion` is the `version` the catalog returns for that type — versions are scoped per type, so a version valid for one agreement is not valid for another.

## Design decisions worth a reviewer's attention

**The endpoint path is unchanged.** `GET /customers/end-user-terms` keeps its path and gains a list response (`operationId` `getEndUserTerms` → `listAgreementDocuments`). Renaming the path would be a second breaking change for no functional gain; happy to rename if you'd prefer to take both at once.

**The catalog is not a requirements list.** Seven supported types does not mean every customer must accept all seven. Which documents are required, for which products and customers, is a legal decision this PR deliberately does not encode.

**`info.version` is intentionally not bumped.** It tracks the published dated version — a live routing prefix, also hardcoded across the docs and the generated client — not each breaking merge. It has not moved since that rule landed in #679, across 36+ `breaking-change` PRs. The `breaking-change` label and API-reviewer approval are the real gate; flagging it here so that decision is explicit rather than implied.

**Clients pinned to the dated version are not affected.** The server change lands this shape on the `/grid/rc` release-candidate surface only, leaving `/grid/2025-10-13` serving the existing contract. Cutting a new dated version is a separate, coordinated release.

**`retrieve_end_user_terms` in `.stainless/stainless.yml` is left alone.** It's a published SDK method name and its path is unchanged, so renaming it is an independent SDK-breaking decision. The three model mappings that pointed at now-deleted schemas are updated.

## Migration for existing consumers

- Acceptance already recorded under the old field maps only to `LIGHTSPARK_END_USER_TERMS`. Acceptance of one agreement never implies another, so the other six must be collected from each customer.
- On update, supplying consents records additional acceptances; evidence already on file for other types is untouched. Omitting the field changes nothing.
- `Customer.agreementConsents` is read-only and returns the most recent acceptance per accepted type. It is omitted entirely until the first acceptance, matching how the old `endUserTermsConsent` behaved.

## Changes: 24 files

- `openapi/components/schemas/customers/Agreement*.yaml` — 6 new schemas (type, acceptance method, consent request/response, document, list response)
- `openapi/components/schemas/customers/EndUserTerms*.yaml` — 3 superseded schemas removed (`EndUserTerms`, `EndUserTermsAcceptanceMethod`, `EndUserTermsConsentRequest`); `EndUserTermsConsent` is kept and marked `deprecated: true`
- `openapi/components/schemas/customers/Customer{,CreateRequest,UpdateRequest}.yaml` — field swapped to the list. On `Customer` the field is **optional and omitted when the customer has no acceptances**, rather than always present as `[]` (per review) — matching the deprecated `endUserTermsConsent`, which already documented omit-when-absent
- `openapi/paths/customers/customers_agreements.yaml` (new) — `GET /customers/agreements`, the catalog of all seven documents
- `openapi/paths/customers/customers_end-user-terms.yaml` — kept and marked `deprecated: true`, still returning its original single-document body so existing callers keep working
- `openapi/webhooks/customer-update.yaml` — 2 examples updated to match `Customer`
- `openapi/components/schemas/errors/Error400.yaml` — `END_USER_TERMS_VERSION_NOT_FOUND` now fires for any of the seven agreements, so its description no longer names the End User Terms specifically. Widened the wording rather than adding an enum value, which would be a wire change for clients to distinguish cases with the same cause and the same remedy
- `.stainless/stainless.yml` — model mappings repointed
- `mintlify/` — disclosures rewritten, snippet + changelog entry, regenerated bundles

## Rebase note

Rebased onto `main` (6 commits, through #934 "bind one funding source per card"). Two conflicts, both resolved keeping each side:

- `Customer.yaml` — main still carried only `endUserTermsConsent`; this branch adds the deprecation marker plus `agreementConsents`. Took this branch's version, which contains both.
- `mintlify/changelog.mdx` — both sides added a new top entry. Kept both, main's card-refunds entry above this one.

Verified after the rebase that main's card work survived intact (e.g. `CardTransaction.originalTransactionId` is present in the rebuilt bundle) alongside the seven-value enum and the deprecated field.

## Test plan

This repo has no test suite; it is spec + docs. Verified:

- `make build` bundles cleanly; both generated bundles regenerated and committed.
- `make lint` exits 0 with **2 warnings / 14 ignored — byte-identical to the `main` baseline**, confirmed by re-running lint on a stashed tree. No new lint problems.
- Parsed the bundle and asserted: the enum holds exactly 7 unique values; all 5 consent fields are required; all three customer schemas expose `agreementConsents` and no longer expose `endUserTermsConsent`; the 200 response resolves to `data[]` of `AgreementDocument` with `{type, version, url}` all required; no `EndUserTerms*` schema remains; `info.version` unchanged.
- An automated review pass caught the stale `.stainless` model mappings, which are fixed here. Its remaining finding is the version-bump question addressed above.

## Public

Consent is now recorded per agreement type, and the agreements endpoint lists every supported document with its current version and URL.

Original PR: https://github.com/lightsparkdev/grid-api/pull/954


openapi/components/schemas/customers/Customer.yaml
```diff
@@ -3,6 +3,7 @@ required:
   - umaAddress
   - platformCustomerId
   - customerType
+  - agreementConsents
```
shreyav: should we make this optional? might not be relevant to all customers

bsiaotickchong: yeah just make it optional bolt. the backend doesn't have to return the agreementConsents in the response if it's empty (update the sparkcore PRs)

shreyav (approved):

PR #997, opened 2026-09-14
Split the spend-limit capability, because issuance and update differ

kphurley7:
## Why

`supportsSpendLimits` was described as covering "creating or updating a card". One boolean cannot answer both, because the two moments genuinely differ:

- **Where Grid makes the authorization decision:** limits are accepted at creation and on update; Grid enforces them in its own decisioning.
- **Where the card issuer makes authorization decisions:** limits are **refused at creation** and accepted on update. Grid refuses with *"set them with `PATCH /cards/{id}` once the card is active"*. On update, Grid writes the limit into the **issuer's own** limit structure rather than enforcing it itself - `maxSpendPerTransaction` becomes their per-transaction purchase cell, `maxSpendPerDay` their daily one. Grid's authorization decisioning never runs for these cards, so a decline still has a single source.

That is not a gap waiting to be closed. The issuer's limits route is **ACTIVE-only**, and a freshly issued card is not yet active, so there is no moment at issuance when a limit could be applied. The distinction is structural and permanent.

## What changed

| | `supportsSpendLimits` | `supportsSpendLimitsAtIssuance` |
|---|---|---|
| Grid decides authorizations | `true` | `true` |
| Issuer decides authorizations | `true` | `false` |

- **`supportsSpendLimits`** - whether a card in this program can have these limits **at all**. Answers "should I offer a spend-limit control for this card", and says where they can only be set after issuance.
- **`supportsSpendLimitsAtIssuance`** - whether they may be supplied on `POST /cards`. This is the discriminating one, and the one a caller cannot otherwise discover: they are building a create request before any card exists.

Create-time field descriptions now reference the issuance flag; update-time ones reference the general flag.

## Why now

Nothing populates `CardCapabilities` yet - the implementation is still an open PR - so there are no consumers and reshaping it costs nothing. That stops being true the moment either lands.

The alternative considered and rejected was scoping the single flag to the create path. It reads wrong on `Card`: the card already exists, the caller wants to know about `PATCH`, and a create-scoped `false` would tell them the opposite of the truth.

## Note for review

`supportsTransactionCountLimit` now carries the same value as `supportsSpendLimitsAtIssuance` for both current programs. That is coincidence rather than redundancy - they answer different questions and can diverge - so they are deliberately not merged.

`info.version` is not bumped: an additive read-only field on a schema nothing populates yet.


kphurley7:
@faraday review

kphurley7:
@greptileai review

pengying:
For issuer authorization decisions, do we want to mix their own auth rules with the limits configured on our platform?  That feels like it could be confusing to debug a decline?

pengying (approved):

PR #1007, opened 2026-09-16
Document the railType filter on GET /discoveries

shreyav:
## Reason

The server has accepted a `railType` query parameter on `GET /discoveries` since the endpoint shipped, filtering the returned rails to one type, but the spec never listed it. Sparkcore is moving its handlers onto query models generated from this spec, and an undeclared parameter would become a 400 the moment that handler switches over. Declaring it here keeps a working filter working and gets it into the docs.

## Overview

One new optional query parameter on `GET /discoveries`, `railType`, with the five values the server recognises: `bank`, `mobile_money`, `upi`, `pix`, `iban`. The `currency` parameter's example, which had been sitting after its schema block, is unchanged in meaning.

## Test Plan

- `make build` rebundles cleanly.
- `redocly lint`: zero errors, same as main.
- `spectral lint --fail-severity=error`: passes.

https://claude.ai/code/session_01KYyhvyVNjMMDrKJqFriCE7

---
_Generated by [Claude Code](https://claude.ai/code/session_01KYyhvyVNjMMDrKJqFriCE7)_


shreyav:
Closing rather than merging. This spec already defines `PaymentRail` (UPPER_SNAKE, 23 values), and adding a second five-value lowercase rail enum on one endpoint would put two incompatible rail vocabularies in the public API. The shape of rail filtering on `/discoveries` needs a decision first; this will come back in whatever form that lands on.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1034, opened 2026-09-18
Cards playground: move Explore flows above Design your card

patcapulong:
## What changed

In the cards playground config panel, the **Explore flows** section now renders above **Design your card**. The two `<section>` blocks in `ConfigurePanel.tsx` swapped places; nothing else changed.

## Why

- The intro copy promises "Issue a Visa debit card and watch the exact API calls fire", but the first control under it was nine rows of design swatches. On a laptop-height viewport the flow tiles sat below the fold.
- No flow depends on the design step. Every action in `data/actions.ts` is `available: () => true`, and clicking a downstream flow fast-forwards a card. The design is live state, so editing it after issuing still updates the card on the phone.
- The wallet playground already orders Explore flows above its deeper configuration (Configure auth), so this brings the two playgrounds in line.
- On mobile (767px and below), the configure view is the landing view and tapping a flow tile jumps to the playground. With the tiles first, that path is above the fixed "Explore playground" button.

## Test plan

- Ran the cards demo locally at `localhost:4002` and confirmed the panel reads intro, Explore flows, Design your card, with the flow grid visible without scrolling.
- Confirmed both Reset buttons still render in their own dividers (Explore flows once a card exists, Design your card once the design differs from the default).
- `tsc --noEmit` reports only pre-existing errors in `AppShell.tsx` and `LiquidGlass.tsx`, unrelated to this change.

Made with [Cursor](https://cursor.com)


patcapulong:
@greptileai

patcapulong:
@faraday review

DhruvPareek (approved):

